PR #5, opened 2017-01-12
cifs: make cifs mount options configurable

dmulder:


ddiss:
Could you change the rapido.conf parameter to CIFS_MOUNT_OPTS instead please.
Otherwise the change looks good - thanks!

PR #11, opened 2017-05-18
runtime.vars: fix dracut --kmoddir parameter

luis-henrix:
The --kmoddir expects the directory for the correct kernel version, not
the base directory where the modules will be installed.

Signed-off-by: Luis Henriques <lhenriques@suse.com>

### DIFF
--- a/runtime.vars
+++ b/runtime.vars
@@ -155,9 +155,10 @@ function _rt_require_dracut_args() {
 	# The optional KERNEL_INSTALL_MOD_PATH rapido.conf parameter can be used
 	# to specify where Dracut should try to pull built kernel modules from.
 	if [ -n "$KERNEL_INSTALL_MOD_PATH" ]; then
-		[ -d "$KERNEL_INSTALL_MOD_PATH" ] \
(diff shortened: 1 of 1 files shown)


runtime.vars
```diff
@@ -155,9 +156,9 @@ function _rt_require_dracut_args() {
 	# The optional KERNEL_INSTALL_MOD_PATH rapido.conf parameter can be used
 	# to specify where Dracut should try to pull built kernel modules from.
 	if [ -n "$KERNEL_INSTALL_MOD_PATH" ]; then
-		[ -d "$KERNEL_INSTALL_MOD_PATH" ] \
+		[ -d "$kmoddir" ] \
 				|| _fail "missing $KERNEL_INSTALL_MOD_PATH"
```
ddiss: Please move the kmoddir definition down so that it's in the if block, and fix the error so that it prints the $kmoddir path.
Looks good otherwise - thanks Luis!

luis-henrix:
Thanks, I've pushed my 'kmoddir_fix' branch again (not sure how to handle it the github way... send a new pull request?).

PR #15, opened 2017-05-26
VM metadata as xattr

ddiss:
The main changes here are to allow Cut scripts to specify
- the CPU and memory requirements for the VM image
- whether or not the VM should be booted with a network


ddiss:
I forgot to add the "support networkless VMs" ticket to the 7f25b77 commit message - https://github.com/ddiss/rapido/issues/8 .

cut_fstests_local.sh
```diff
@@ -38,3 +38,4 @@ dracut  --install "tail blockdev ps rmdir resize dd vim grep find df sha256sum \
 	$DRACUT_OUT || _fail "dracut failed"
 
 _rt_xattr_vm_networkless_set "$DRACUT_OUT"
+_rt_xattr_vm_resources_set "$DRACUT_OUT" "2" "2048M"
```
luis-henrix: I guess this could be also added to the cut_fstests_ceph.sh ;-)

luis-henrix:
I didn't manage to do a full review of this massive patchset, but I couldn't find real problem (using attrs for setting config values is... odd, but meh :-)

Anyway, I've also done a quick test and it seems to work fine in my environment.  So... feel free to merge this.  Thanks!

cut_fstests_local.sh
ddiss: This is only really needed to accommodate for the 1G zram disks that are used for fstests_local. If you're hitting the oom-killer while running fstests_ceph, then feel free to propose a change :).

Issue #6, opened 2017-04-10
remove hardcoded library paths


ddiss:
Rapido cut_ scritps that install binaries with dlopen()'ed runtime library dependencies need to explicitly direct Dracut to install such dependencies. This is currently done using an absolute path (based on openSUSE install path), which works on SUSE based platforms, but fails on others (Debian, etc.).

Instead of using an absolute path, dlopen()'ed library dependencies should be looked up, based on /etc/ld.so.conf configuration. This could potentially be done with ldconfig -p.

PR #17, opened 2017-05-31
PR remove hardcoded library paths

luis-henrix:
This is an RFC to fix PR #6 
I haven't thoroughly tested all the cut_* scripts changes, but I believe they should be OK.
Please let me know what you think about the approach.

### DIFF
--- a/cut_cephfs.sh
+++ b/cut_cephfs.sh
@@ -17,13 +17,14 @@ RAPIDO_DIR="$(realpath -e ${0%/*})"
 
 _rt_require_ceph
 _rt_require_dracut_args
+_rt_require_lib "libkeyutils.so.1 libhandle.so.1 libssl.so.1"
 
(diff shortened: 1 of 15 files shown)


luis-henrix:
Thank you for your review, David.  I've re-pushed my pr_hardcoded_lib_paths branch with the changes you suggested (and I was shocked to see that I didn't used tabs for indentation!!!).

ddiss:
One thing I've noticed is that I now may 32bit and 64bit libs for a single entry, e.g.:
```
_rt_require_lib "libkeyutils.so.1...
sbin/ldconfig -p | awk "/libkeyutils.so.1/ {print}"
libkeyutils.so.1 (libc6,x86-64) => /lib64/libkeyutils.so.1 libkeyutils.so.1 (libc6) => /lib/libkeyutils.so.1
```

Still, I don't think it's a big issue, so will go ahead and merge this - thanks Luis!

PR #19, opened 2017-06-01
runtime.vars: expect root repository path for CEPH_SRC

ddiss:
Until now, the CEPH_SRC rapido.conf parameter has required a path to
the "src" subdirectory of the Ceph source repository. This is counter
intuitive, given that all other X_SRC rapido.conf parameters correspond
repository root paths.

Drop the CEPH_SRC "src" subdirectory path requirement, and add temporary
support for translating CEPH_SRC="$ceph_repo/src" configurations into
into CEPH_SRC="$ceph_repo/src/../".

### DIFF
--- a/rapido.conf.example
+++ b/rapido.conf.example
@@ -85,12 +85,13 @@ HOSTNAME2="rapido2"
 #############################
 
 ##### Ceph globals #####
-# Path to src directory within Ceph source repository. Only required if a Ceph
-# specific Cut script will be run. vstart.sh should be run prior to Cut.
+# CEPH_SRC should correspond to a checkout and build of the Ceph source
+# available at https://github.com/ceph/ceph. Only required if a Ceph specific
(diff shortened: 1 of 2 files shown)


luis-henrix:
This makes perfect sense, and the patch looks correct to me!  I actually remember scratching my head about this when I first tried rapido, but then completely forgot about it :-)
TBH, I would also be OK to simply modify the meaning of CEPH_SRC without respecting the "ABI change", failing with a meaningful message in legacy configs.  But I understand your preference.  FWIW:

Acked-by: Luis Henriques <lhenriques@suse.com>

ddiss:
Thanks Luis! I'll raise a separate bug to track removal of the legacy CEPH_SRC and FSTESTS_DIR behaviour. It clutters up the source, so I hope to get rid of it in a week or so, by which time everyone should have fixed there configs :)

PR #26, opened 2017-06-27
Missing dependencies for xfstests

luis-henrix:
I would like to add these extra 2 dependencies for running xfstests in rapido:
- fallocate is used in several generic tests in the test-suite
- libgdbm is required to compile (and run) tests that need src/dbtest.c

### DIFF
--- a/cut_fstests_cephfs.sh
+++ b/cut_fstests_cephfs.sh
@@ -18,7 +18,7 @@ RAPIDO_DIR="$(realpath -e ${0%/*})"
 _rt_require_dracut_args
 _rt_require_ceph
 _rt_require_fstests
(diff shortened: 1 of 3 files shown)


ddiss:
I haven't been able to spot any tests using the fallocate command - all tests seem to be using xfs_io, or fsx/fsstress, etc. Am I missing something?

luis-henrix:
Rats! You're right... I added it to my cut_* scripts so long ago that I forgot why.  Probably to run some manual tests, but definitely not as an xfstests dependency.
Oh well, I'll just wipe it.  What about the other dependency?  It looks like only generic/010 is using dbtest.

ddiss:
The other dependency fix looks correct. Please force push to this branch with your gdbm library fix only.

luis-henrix:
Done!

PR #30, opened 2017-07-13
Small rapido improvements

mwilck:
I finally got started with rapido, and here is a first set of minor fixes matching my workflow. Please check if you're ok with it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 rapido.conf
 initrds/myinitrd
+\#*
+*~
+*.pid
--- a/cut_blktests_rbd.sh
+++ b/cut_blktests_rbd.sh
@@ -19,7 +19,7 @@ _rt_require_dracut_args
 _rt_require_ceph
 _rt_require_blktests
 
-dracut  --install "tail blockdev ps rmdir resize dd vim grep find df sha256sum \
+"$DRACUT" --install "tail blockdev ps rmdir resize dd vim grep find df sha256sum \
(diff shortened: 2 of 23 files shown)


mwilck:
@morbidrsa, yes I had dracut problems. bsc#1048606 for example. Also, I'm wondering if I could use this for my pending dracut/iSCSI boot fixes which will definitely require using different dracut code than installed on my workstation.
That touches a generic issue I'm having with rapido, it's really well suited for quickly testing kernel changes, but I'd also consider using it for testing user space stuff like udev, multipathd etc. That would require fetching these from other places than the standard installation path, too. 

Wrt the kernel config, there must be a reason why it's different from stock. I didn't realize that these configs existed in the first place. I believe it's for speed - compiling the full SLES kernel requires a lot of patience. My change was intended to make sure that rapido also works with the stock config, not to replace rapido's internal one.

ddiss:
@mwilck thanks for the changes - they look fine to me, one request though:
For testing of user space programs from source (e.g. blktests), we've used X_SRC rapido.conf parameters until now. Would the same work for your Dracut use case? E.g

<pre>
rapido.conf
  DRACUT_SRC=/my/dracut/repo

function _rt_require_dracut_args() {
  ...
  if [ -n "$DRACUT_SRC" ]; then
    DRACUT="$DRACUT_SRC/dracut.sh"
    ...
  else
    DRACUT="dracut"
  fi
</pre>

mwilck:
@ddiss: Thanks for the review. I updated the pull request according to your scheme.

PR #32, opened 2017-07-17
Misc. mpath network, zram_hot_add.sh and Ceph udev path changes

ddiss:

PR #33, opened 2017-07-17
rapido / multipath fixes and enhancements

mwilck:


cut_mpath_local.sh
```diff
@@ -34,9 +34,9 @@ _rt_require_dracut_args
 	--include "$RAPIDO_DIR/mpath_local_autorun.sh" "/.profile" \
 	--include "$RAPIDO_DIR/rapido.conf" "/rapido.conf" \
 	--include "$RAPIDO_DIR/vm_autorun.env" "/vm_autorun.env" \
-	--add-drivers "virtio_scsi" \
-	--modules "bash base network ifcfg systemd systemd-initrd \
-		   dracut-systemd multipath" \
+	--add-drivers "virtio_scsi virtio_pci sd_mod" \
+	--kernel-cmdline "root=/dev/sda rd.timeout=5" \
```
ddiss: hmm, I'd prefer to avoid any use of "--kernel-cmdline root" - mpath_local images already boot fine for me on oS Leap 42.2. What specific release are you running on?

cut_mpath_local.sh
```diff
@@ -34,9 +34,9 @@ _rt_require_dracut_args
...
+	--modules "bash base systemd systemd-initrd dracut-systemd multipath" \
```
ddiss: FWIW, I've already queued a matching change for this: https://github.com/rapido-linux/rapido/pull/32/commits/eb5b439304cc8ae4ca775d1f5403efcdc36d20be

runtime.vars
```diff
@@ -265,3 +272,21 @@ function _rt_xattr_vm_resources_get
...
+		return
+	}
+
+	_RT_MPATH_DIR=$(mktemp -d /tmp/rapido.XXXXXX)
+	[ -n "$_RT_MPATH_DIR" ] || _fail "failed to create tmp dir"
+	_rt_push_unwind "rm -rf $_RT_MPATH_DIR"
+
+	make -C "$MULTIPATH_SRC" DESTDIR="$_RT_MPATH_DIR" \
+	     ${MULTIPATH_MAKEFLAGS} install || \
+		_fail "failed to install multipath-tools from $MULTIPATH_SRC"
```
ddiss: I want to avoid getting Rapido involved in building projects.
Please just assert that the source at $MULTIPATH_SRC is already compiled, and then just use '--include "$MULTIPATH_SRC" /X' in the cut script.
This is similar to what's already done for FSTESTS_SRC and BLKTESTS_SRC.

ddiss: Also, if you only need a few things from the compiled "$MULTIPATH_SRC", then consider installing them (via dracut --install) separately, rather than dumping the full directory in the initramfs.

cut_mpath_local.sh
mwilck: Hmhmhm, suddenly it works without "root=" for me too. I swear I fought all afternoon to get this up and only root=worked. Need to check again.

runtime.vars
mwilck: That's difficult. A multipath installation contains quite some files. I don't think it's reasonable to include the whole source tree, and cherry picking only the absolutely needed stuff creates a long hand-coded file list that might change with the next package release. And I need to make sure all files that dracut installs from the system package are overwritten (AFAICS you don't have that problem for FSTESTS_SRC).

I really think this is what "make install" was created for.

cut_mpath_local.sh
mwilck: I see. didn't check the "misc" branch.

runtime.vars
mwilck: '--include "$MULTIPATH_SRC" /X' definitely doesn't work because the directory structure in the source tree is different from the installed code base. Picking individual files is hard because I need to re-implement parts of the Makefile logic.
Please consider this again, I have no better idea than "make install".

cut_mpath_local.sh
mwilck: Reproduced again. This happens if I work without "root=" argument: [screenlog.txt](https://github.com/rapido-linux/rapido/files/1154084/screenlog.txt).

This is TW, dracut-044.1-1.1.x86_64, SLE12-SP3 kernel 4.4.73 with stock SLE12-SP3 kernel configuration.

mwilck: The same thing happens to me also if I use the SLE12-SP2 dracut code. And it also happens for the "nvme_rdma" case, which is driving me nuts because I haven't found a workaround there like I did for the mpath_local case.
Any idea what I'm doing wrong?

runtime.vars
ddiss: In that case, how about a MULTIPATH_INSTALL_DESTDIR rapido.conf option, which is then used as the dracut install source?
If you'd still like to automate the multipath source build, then I'd recommend doing it from a parent git repository with a git submodule link to the rapido core repo. See https://github.com/ddiss/rapido-utils - IMO it relatively nicely keeps this kind of clutter out of the core repo.

cut_mpath_local.sh
ddiss: Strange. I'll need to revisit this on my local setup, but IIRC I had issues with dropping to a prompt from systemd before adding the rd.systemd.unit=emergency kernel parameter, after that it worked without issues.
It might also be worth trying one of the kernel configs from the rapido repo.

runtime.vars
mwilck: MULTIPATH_INSTALL_DESTDIR should work, thanks for the idea.

mwilck:
This is superseded by https://github.com/ddiss/rapido/pull/1. I'm still working on the revised MULTIPATH_SRC patch.

mwilck:
I've added the MULTIPATH_DESTDIR stuff in a new branch: mwilck/rapido@290d37d464dd079402f2880f33412e4409b59196

mwilck:
I've added the MULTIPATH_DESTDIR stuff in a new branch: mwilck/rapido@290d37d464dd079402f2880f33412e4409b59196

mwilck:
I've added the MULTIPATH_DESTDIR stuff in a new branch: mwilck/rapido@290d37d464dd079402f2880f33412e4409b59196

PR #37, opened 2017-12-20
minor miscellaneous changes

ddiss:


luis-henrix:
It looks good to me!  One of these days we should probably cleanup all the cut_* scripts so that they don't dup all that info.  But yeah that'll take a little bit of time ;-)

luis-henrix:
(i will never learn how this github thing works!)

PR #43, opened 2018-05-04
Simple example scripts and readme improvements

ddiss:


ddiss:
FWIW, a rendering of the asciidoc documentation can be found at:
https://github.com/ddiss/rapido/blob/simple_example/readme.adoc

ddiss:
I've pushed a couple of extra documentation changes which update vstart_kernel_test.txt and move it under a new doc subdir: https://github.com/ddiss/rapido/blob/simple_example/doc/ceph_vstart_kernel_client_guide.adoc

luis-henrix:
(sorry for the delay reviewing)
This looks good to me -- the updated documentation is always welcome, and I guess the template cut/autorun scripts are a very good idea.  Thanks @ddiss

luis-henrix:
Why do I always do this?  grrr

PR #47, opened 2018-05-18
Fix for dracut.conf.d processing and fstests_cifs_autorun guard

ddiss:


dracut.conf.d/01-rapido-dracut.conf
```diff
@@ -0,0 +1,9 @@
+# disable compression for performance
```
ddiss: It probably makes sense to add a note here that this file is the wrong place for Rapido configuration (see ../rapido.conf)

ddiss: fixed via forced push...

PR #51, opened 2018-07-24
CTDB + Samba + libcephfs cut / autorun scripts

ddiss:
The three-VM-support changes might be interesting to others here. I'm still not happy with the rapido.conf VM network syntax, but couldn't think of anything nicer, aside from perhaps moving to json based configuration in future.

### DIFF
--- a/tools/br_setup.sh
+++ b/tools/br_setup.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright (C) SUSE LINUX GmbH 2016, all rights reserved.
+# Copyright (C) SUSE LINUX GmbH 2016-2018, all rights reserved.
 #
 # This library is free software; you can redistribute it and/or modify it
 # under the terms of the GNU Lesser General Public License as published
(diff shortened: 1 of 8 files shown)


tools/br_setup.sh
```diff
@@ -35,39 +35,62 @@ if [ -n "$BR_IF" ]; then
...
-unwind="ip link set $TAP_DEV0 nomaster; ${unwind}"
-echo "+ created $TAP_DEV0"
-
-ip tuntap add dev $TAP_DEV1 mode tap user $TAP_USER || exit 1
-unwind="ip tuntap delete dev $TAP_DEV1 mode tap; ${unwind}"
-ip link set $TAP_DEV1 master $BR_DEV || exit 1
-unwind="ip link set $TAP_DEV1 nomaster; ${unwind}"
-echo "+ created $TAP_DEV1"
+for vm_num in `seq 1 "$VM_MAX_COUNT"`; do
+	eval mac_addr='$MAC_ADDR'${vm_num}
```
morbidrsa: Can you make MAC_ADDR an array? So we can access it as mac_addr = ${MAC_ADDR[$vm_num]}.

tools/br_setup.sh
```diff
@@ -35,39 +35,62 @@ if [ -n "$BR_IF" ]; then
...
+	eval tap='$TAP_DEV'$((vm_num - 1))
+	[ -z "$tap" ] \
+		&& _fail "TAP_DEV$((vm_num - 1)) not configured"
+
+	_br_tap_setup "$tap"
+
+	[ -z "$BR_DHCP_SRV_RANGE" ] \
+		&& continue	# no DHCP server
+
+	eval ip='$IP_ADDR'${vm_num}
```
morbidrsa: Same for ip

tools/br_setup.sh
```diff
@@ -35,39 +35,62 @@ if [ -n "$BR_IF" ]; then
...
+	[ -z "$ip" ] \
+		&& continue	# no VM IP assigned for DHCP
+
+	# add an explicit DHCP server host entry for this VM
+	dhcp_hosts="$dhcp_hosts --dhcp-host=${mac_addr},${ip}"
+
+	eval hn='$HOSTNAME'${vm_num}
```
morbidrsa: And hostname

tools/br_teardown.sh
```diff
@@ -29,15 +29,20 @@ if [ -n "$BR_DHCP_SRV_RANGE" ]; then
...
 fi
 
-ip link set dev $TAP_DEV1 down || exit 1
-ip link set dev $TAP_DEV0 down || exit 1
-ip link set dev $BR_DEV down || exit 1
-
-ip link set $TAP_DEV1 nomaster || exit 1
-ip tuntap delete dev $TAP_DEV1 mode tap || exit 1
+for vm_num in `seq 1 "$VM_MAX_COUNT"`; do
+	eval mac_addr='$MAC_ADDR'${vm_num}
```
morbidrsa: Here as well

tools/br_teardown.sh
```diff
@@ -29,15 +29,20 @@ if [ -n "$BR_DHCP_SRV_RANGE" ]; then
...
+	if [ -z "$mac_addr" ]; then
+		continue
+	fi
+	eval tap='$TAP_DEV'$((vm_num - 1))
```
morbidrsa: And here

tools/br_setup.sh
ddiss: Thanks for the feedback.

Do you mean changing the rapido.conf VM network configuration syntax to use an array?
eval x='Y'${vm_num} is currently used to programatically obtain the rapido.conf network parameter for a given vm index. It's ugly, but short of moving to ini or yaml format I'm not aware of a better way.

ddiss: Oh btw, x='Y'${vm_num} is already used in _vm_start() , so this loop is mostly copying what's already done there.

morbidrsa: Damn, ok. Haven’t realized it comes from rapido.conf. I guess that’s not something we can change easily then.

ddiss: Yeah. I'd like to change the VM network config syntax in future, but the new format would need to be properly defined / reviewed separately first.

ddiss:
Any other feedback on these changes? I'm still using these downstream and haven't run into any issues.

ddiss:
Ping? Would still like to get these changes in.
@morbidrsa if you're against adding 3-vm support, then I could rework the ctdb_cephfs changes to only expect two VMs maximum.

ddiss:
Superseded by https://github.com/rapido-linux/rapido/pull/56 .

PR #53, opened 2018-08-22
rbd-nbd

ddiss:


ddiss:
This change doesn't conflict with the other 3VM + CTDB changes, so submitting separately.

PR #54, opened 2018-08-30
lrbd test support

ddiss:


ddiss:
Minor update:
- fix some whitespace issues
- add rapido.conf.example entries for lrbd

morbidrsa:
Looks good, but do we really want to go down the 100MB initrd road? I guess yes otherwise you wouldn’t have implemented it.

ddiss:
> Looks good, but do we really want to go down the 100MB initrd road? I guess yes otherwise you wouldn’t have implemented it.

Given that the bloat is isolated to this image (due to Ceph and Python), I'm okay with it.
The lrbd backend offers an environment much closer to what customers use for LIO+RBD, compared to the minimal lio_rbd backend.

morbidrsa:
Ok then

PR #57, opened 2018-11-09
openiscsi: add cut and autorun

ddiss:
Generate an image with iscsid and iscsiadm from OPENISCSI_SRC. On boot,
connect to the target at INITIATOR_DISCOVERY_ADDR, issue a SendTargets
request and then login to any discovered targets.

### DIFF
--- a/cut_openiscsi.sh
+++ b/cut_openiscsi.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+#
+# Copyright (C) SUSE LINUX GmbH 2018, all rights reserved.
+#
+# This library is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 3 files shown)


ddiss:
@gonzoleeman : Might be of interest.

gonzoleeman:
Yes, very interesting.

ddiss:
@gonzoleeman can I take that as a review or ack? :-)
If so, I'll merge it now as-is.

gonzoleeman:
David: the code *looks* correct to me (the iscsi parts that I grok), and I _assume_ you've tested it (since I know you, so I know you normally would). Given that, sure, call that an Ack.

PR #59, opened 2018-12-18
fstests: add Btrfs specific cut + autorun

ddiss:


ddiss:
I've reworked the patchset to use dracut --install for xfstest/src binaries, so that shared-object dependencies are pulled in.

ddiss:
I unfortunately submitted the wrong (non-v2) branch for my rework, sorry. Will rebase the changes and resubmit.

PR #61, opened 2018-12-19
Fstests exclude

morbidrsa:
Provide a way to exclude xfstests test in rapido


morbidrsa:
Yep you're right. Withdrawing the PR.

PR #64, opened 2018-12-20
rapido: consolidate all scripts into one command

morbidrsa:
Consolidate all scripts into one command, so the user only calls 'rapido'
with an appropriate sub-command, like 'rapido start-vm' to start the VM or
'rapido setup-network' to setup the brides, etc. for the VM network.

Signed-off-by: Johannes Thumshirn <jthumshirn@suse.de>


morbidrsa:
As a follow up cleanup, we could move all the cut_ scripts into their own folder, like you did with the autorun scripts in #60.

morbidrsa:
The cut, boot and list part is a no-brainer, will be doing that. I have an idea for the short description and help messages, so I'll work that in as well.

For the networking sub-subcommand this sounds good, but I don't know a good way how to solve it yet, I'll have to think about it a bit more.

PR #72, opened 2019-02-11
RFC: Fix New ceph.conf format parsing #71

### LINKED ISSUE #71: New ceph.conf format parsing
Nautilus seems to have slightly changed the ceph.conf format and it causes _ini_parse function to fail parsing it.  The change in question is regarding the monitors section, which can now have a list of IP addresses:
```
 [mon.a]
        mon host =  [192.168.155.1:40880]
```
The '[<addr>:<port>]' is seems to be confusing the parsing code. The same occurs for the global section:
```
[global]
        mon host =  [v2:192.168.155.1:40879,v1:192.168.155.1:40880]
```
I've hacked some code to parse this new config using awk, which should be a bit easier to maintain IMO.  Here's what I currently have:
```
function _ini_parse() {
	local ini_file=$1
	local ini_section=$2
	shift 2
	local ini_key=\"$@\"

	awk -v section="^\\\[$ini_section\\\]" \
		-v key="$ini_key" \
		'$1 ~ section {
			got_section = 1;
			next
		}
		/^\[.*\]/{
			got_section = 0;
			got_key = 0;
		}
		$0 ~ key {
			got_key = 1;
		}
		got_section && got_key && NF {
			print substr($0, index($0, "=") + 1)
			exit 0;
		}' $file
}
```
Would something this be acceptable?  The idea is to actually wrap _ini_parse() with some more high-level functions (such as ceph_get_key, ceph_get_mon_addr...).

luis-henrix:
This RFC tries to fix #71 (although the 1st commit is actually unrelated...).  This is just an RFC as I didn't actually did full testing to all the scripts and there are probably several inconsistencies (e.g. functions may deserve better names :-) ).
Anyway, I'll review this again myself later, but wanted to have something pushed out early in case someone hits the same issue.

### DIFF
--- a/autorun/blktests_rbd.sh
+++ b/autorun/blktests_rbd.sh
@@ -18,6 +18,7 @@ if [ ! -f /vm_autorun.env ]; then
 fi
 
 . /vm_autorun.env
+. /vm_ceph.env
 
 set -x
 
--- a/autorun/cephfs.sh
+++ b/autorun/cephfs.sh
@@ -18,6 +18,7 @@ if [ ! -f /vm_autorun.env ]; then
 fi
 
 . /vm_autorun.env
+. /vm_ceph.env
 
 set -x
 
@@ -29,19 +30,8 @@ cat > /etc/hosts <<EOF
 127.0.0.1	$hostname_fqn	$hostname_short
 EOF
 
-_ini_parse "/etc/ceph/keyring" "client.${CEPH_USER}" "key"
-[ -z "$key" ] && _fatal "client.${CEPH_USER} key not found in keyring"
-if [ -z "$CEPH_MON_NAME" ]; then
-	# pass global section and use mon_host
-	_ini_parse "/etc/ceph/ceph.conf" "global" "mon_host"
-	MON_ADDRESS="$mon_host"
-else
-	_ini_parse "/etc/ceph/ceph.conf" "mon.${CEPH_MON_NAME}" "mon_addr"
-	MON_ADDRESS="$mon_addr"
-fi
-
 _vm_ar_dyn_debug_enable
 
 mkdir -p /mnt/cephfs
-mount -t ceph ${MON_ADDRESS}:/ /mnt/cephfs -o name=${CEPH_USER},secret=${key}
+mount -t ceph ${CEPH_MON_ADDRESS_V1}:/ /mnt/cephfs -o name=${CEPH_USER},secret=${CEPH_USER_KEY}
 set +x
--- a/autorun/ctdb_cephfs.sh
+++ b/autorun/ctdb_cephfs.sh
@@ -18,6 +18,7 @@ if [ ! -f /vm_autorun.env ]; then
 fi
 
 . /vm_autorun.env
+. /vm_ceph.env
 
 _vm_ar_dyn_debug_enable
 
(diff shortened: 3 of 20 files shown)


ddiss:
The first change looks fine (feel free to PR separately if you like).
I'll need a little more time to look over the others - I'm a little unsure about the ini-file wrappers, as it makes it a little harder to determine which magic env vars will be filled by looking at the caller.

luis-henrix:
> The first change looks fine (feel free to PR separately if you like).
> I'll need a little more time to look over the others - I'm a little unsure about the ini-file wrappers, as it makes it a little harder to determine which magic env vars will be filled by looking at the caller.

Yep, sure.  I'm not sure myself if that's the best approach.  TBH I started by having a different function for each parameter and having them returned by the function instead of setting a variable.  That's probably a better approach.

ddiss:
> Yep, sure.  I'm not sure myself if that's the best approach.  TBH I  started by having a different function for each parameter and having  them returned by the function instead of setting a variable.  That's  probably a better approach.

Sounds interesting, feel free to PR :)

luis-henrix:
Ok, I've (force!) pushed a new version without any variables polluting the environment.  Still RFC, not fully tested (specially the changes ctdb/samba changes).

luis-henrix:
Thanks for the review.  Actually, after some more tests, I had already a couple of changes queued locally.  (I know `ceph-conf` is a bit too big (~10M here) but I'm starting to wonder if we shouldn't simply include  it in the image and use it to do the parsing...)

Anyway, I had already detected the `$ini_file` issue (not sure how that one slipped).  All the other changes you propose seem to make sense too.

A big issue I found is actually the usage of v1/v2 in the monitors addresses.  It looks like the old format is being deprecated and the address is now in the [global] section with something like:
```
mon host =  [v2:192.168.155.1:40212,v1:192.168.155.1:40213]
```
I was planning to simply pick the v1 address and discard everything else by running the value returned by `_ini_parse` through:
```
sed -e 's/[][]//g' | sed -E 's/.*v1:([0-9.:]*).*/\1/g'
```
Would this be acceptable?  Probably in the future we'll want to have a version that would use v2 as well.

ddiss:
Gah, I just noticed (in a master vstart) that the "mon host" key is defined in a second "[global]" section, so the "leaving section of interest" early-exit isn't an option...

luis-henrix:
I've just pushed another version of the fix for this issue.  This time I've added a new function that uses ceph-conf to generate a file with the ceph configs ready to be source by rapido scripts.  I've just pushed it as a PoC, as only the fstest_cephfs scripts have been converted to use it.  It seems to be working, but there are still a few things:

- The monitor address will be the msgr v1 address, the v2 (if present) will be ignored.
- The config file is currently being deleted after being copied into the initrd, and the 'cut' operation will fail if the file exists.  The alternatives: 1) always overwrite the file or 2) use mktmp to generate a temporary file, and never delete it.

@ddiss If this approach is acceptable, I can work on migrating all the other scripts to use it (and drop the _ini_parse code).

ddiss:
@luis-henrix looks nice :) A couple of minor nits...
- please add a file path parameter so that cut scripts control where it goes (mktemp)
  + rename it _rt_write_ceph_config() and fail if the path doesn't exist
  + VM_CEPH_CONF can then be dropped
- use plain bash for the mon_addr string manipulation, i.e.
```
# Nautilus addr specifies two messenger endpoints, pull out v1 only
value="${value##*v1:}"
value="${value%%]}"
```
  + use CEPH_MON_ADDR_V1 (might also be worth adding CEPH_MON_ADDR_V2 to the conf)
- use consistent spacing/underscores for "mds root ino uid" and "mds_root_ino_gid"

luis-henrix:
And here's another revision of my patchset, just before going on weekend :-)
I've modified all the scripts that require the new config parser, either directly or indirectly (i.e those that were calling _vm_ar_rbd_map).
I haven't tested all these scripts, so I may have broken something.  I'll re-review everything again later or, more likely, on Monday.

ddiss:
Looks good to me - thanks a lot for adding this, Luis!

PR #74, opened 2019-02-14
Drop ceph udev parttypeuuid and add simple cifs mount runner

ddiss:


ddiss:
@luis-henrix this shouldn't collide with your ceph.conf parsing changes.

PR #76, opened 2019-02-15
Minor vm_ceph.env follow-ups

ddiss:


ddiss:
> The only thing that may be worth changing in these patches is to also drop some extra "local value" I left in function 

Thanks, I've pushed an update with the extra "local value=" -> "value=" change squashed in. Please merge.

PR #77, opened 2019-02-18
Drop dracut network and ifcfg modules

ddiss:
This trims most network attached images by 15-20% . I've tacked on a few minor cephfs cut/autorun changes.


ddiss:
Thanks Luis! merging...

PR #80, opened 2019-04-03
samba_kernel_cephfs: kernel-backed Samba share runner

ddiss:
This mostly matches samba_cephfs, but uses a kernel CephFS mountpoint
instead of libcephfs for filesystem I/O.


ddiss:
> This all looks good to me -- it looks similar enough to the samba_cephfs scripts, although I have to confess that I haven't really tested it (I guess I need to start thinking about setting up both an SMB and an NFS test environment).

Thanks for the feedback. I've tested locally, so don't expect it as part of review - merging...

PR #85, opened 2019-04-25
Add support for LTP

frankenmichl:
The Linux Test Project [1] is a collection of tools for testing the
Linux kernel and related features.
These tests can/will harm the system you run them on, so running them
inside virtual machines is neccessary for developing new testcases.

Rapido can build a ramdisk image with the complete LTP binaries and
execute a single test or the included testrunner.

[1] https://github.com/linux-test-project/ltp/

Signed-off-by: Michael Moese <mmoese@suse.de>

### DIFF
--- a/autorun/ltp.sh
+++ b/autorun/ltp.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+#
+# Copyright (C) SUSE LINUX GmbH 2019, all rights reserved.
+#
+# This library is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published
+# by the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) version 3.
+#
+# This library is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public
+# License for more details.
+
+
+if [ ! -f /vm_autorun.env ]; then
+	echo "Error: autorun scripts must be run from within an initramfs VM"
+	exit 1
+fi
+
+. /vm_autorun.env
+
+LTP_DIR="/opt/ltp"
+[ -d "$LTP_DIR" ] || _fatal "LTP missing"
+
+_vm_ar_dyn_debug_enable
+
+set -x
+
+echo "running LTP Test $LTP_AUTORUN_CMD" 
+#
+if [ -n "$LTP_AUTORUN_CMD" ]; then
+        cd ${LTP_DIR} || _fatal
+        eval "$LTP_AUTORUN_CMD"
+fi
+
--- a/cut/ltp.sh
+++ b/cut/ltp.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+#
+# Copyright (C) SUSE LINUX GmbH 2019, all rights reserved.
+#
+# This library is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published
+# by the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) version 3.
+#
+# This library is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public
+# License for more details.
+
+RAPIDO_DIR="$(realpath -e ${0%/*})/.."
+. "${RAPIDO_DIR}/runtime.vars"
+
+_rt_require_dracut_args
+_rt_require_conf_dir LTP_BIN
+
+"$DRACUT" \
+	--install "ps rmdir dd mkfs.xfs" \
+	--include "$RAPIDO_DIR/autorun/ltp.sh" "/.profile" \
+	--include "$RAPIDO_DIR/rapido.conf" "/rapido.conf" \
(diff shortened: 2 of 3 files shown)


cut/ltp.sh
```diff
@@ -0,0 +1,32 @@
...
+#
+# This library is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public
+# License for more details.
+
+RAPIDO_DIR="$(realpath -e ${0%/*})/.."
+. "${RAPIDO_DIR}/runtime.vars"
+
+_rt_require_dracut_args
```
ddiss: Please add "_rt_require_conf_dir LTP_SRC" here

cut/ltp.sh
```diff
@@ -0,0 +1,32 @@
...
+
+"$DRACUT" \
+	--install "ps rmdir dd mkfs.xfs" \
+	--include "$RAPIDO_DIR/autorun/ltp.sh" "/.profile" \
+	--include "$RAPIDO_DIR/rapido.conf" "/rapido.conf" \
+	--include "$RAPIDO_DIR/vm_autorun.env" "/vm_autorun.env" \
+	--include "$LTP_SRC" "/opt/ltp"  \
```
ddiss: Doesn't ltp include a bunch of (precompiled) binaries? I think it makes sense to behave similar to cut/fstests_xfs.sh here:
- copy the $LTP_SRC directory into the same path in the image
- use ``--install`` for binaries to ensure that the corresponding shared library dependencies get pulled in, e.g.
```
--install "ps rmdir dd mkfs.xfs ${LTP_SRC}/bin/*" \
...
--include "$LTP_SRC" "$LTP_SRC"  \
...
```

rapido.conf.example
```diff
@@ -227,3 +227,10 @@ INITIATOR_IQNS="iqn.2007-10.com.github:sahlberg:libiscsi:iscsi-test \
 # when issuing a SendTargets discovery request.
 #INITIATOR_DISCOVERY_ADDR=""
 ################################
+
+########## cut/ltp.sh ##########
+# LTP_SRC should correspond to the ltp binary tree after make install 
+#LTP_SRC=
```
ddiss: ``make install`` or just ``make``? Rapido normally runs using (compiled) source repositories, rather than installed test suites, so that'd be preferred here too.

autorun/ltp.sh
```diff
@@ -0,0 +1,36 @@
...
+
+if [ ! -f /vm_autorun.env ]; then
+	echo "Error: autorun scripts must be run from within an initramfs VM"
+	exit 1
+fi
+
+. /vm_autorun.env
+
+LTP_DIR="/opt/ltp"
+[ -d "$LTP_DIR" ] || _fatal "LTP missing"
```
ddiss: No need to set LTP_DIR here if you use the same rapido.conf path (see below)

cut/ltp.sh
frankenmichl: Makes sense. I'll add that.

cut/ltp.sh
frankenmichl: The default binary location for LTP is /opt/ltp - so I think it is convenient to put it there in the image as well. Unfortunately, I would not just include the binaries, because we would miss the testrunner,  the groups, etc.

rapido.conf.example
frankenmichl: Well, usually we really need to do `make install`, because the LTP build system places the binaries in the same location as the sources. So we don't have a dedicated directory for the build results. We don't have to install it to some specific location in the system.
Usually I just do `./configure --prefix=~/build/ltp` and get a clean tree of binaries for my testcases. 
I think, unless someone wants to change the LTP Makefiles, we should go with this. 

Anyway, I should mention the prefix in the comment.

ddiss: Okay, if you want to have ltp installed from the local system, rather than a source directory, then please drop the ``LTP_SRC`` parameter and just hardcode the default install path ("/opt/ltp").

cut/ltp.sh
ddiss: With ``LTP_SRC`` being dropped, please use something like the following instead:
```
LTP_DIR="/opt/ltp"
[ -d "$LTP_DIR" ] || _fail "LTP missing"
```
``LTP_DIR`` can then also be used for ``--install`` and ``--include``

ddiss:
@frankenmichl : any update on this patchset? I'd like to get it merged with the minor changes suggested.

cut/ltp.sh
frankenmichl: We don't really dependencies in LTP - the standard library will be there anyway I guess. However, if I would use /opt/ltp, on typical systems you would need superuser permission, which I would prefer to avoid.

frankenmichl:
I'm not sure how to proceed with your remarks regarding the LTP_DIR. I'd rather not require permission to write to /opt when developing/running testcases..

ddiss:
@frankenmichl : mind if I push with the following change squashed in?
https://github.com/ddiss/rapido/commit/ca049179d71b5cdc9ec24ce3b7979ae934e2a91a 

>  I'm not sure how to proceed with your remarks regarding the  LTP_DIR. I'd rather not require permission to write to /opt when  developing/running testcases.

This seems to be a misunderstanding, as I wasn't suggesting writing to /opt on the base system.

frankenmichl:
I'm fine with your changes - and I tested your suggestion.

ddiss:
> I'm fine with your changes - and I tested your suggestion.

Pushed to master as 1783f67d7c85ba89005059fb942493e178321889 - thanks!

PR #88, opened 2019-06-13
tcmu_zbc_loop: tcmu-runner Zone Block Commands (ZBC) device

ddiss:
The tcmu_zbc_loop VM image is cut with the LIO tcmu kernel module and
tcmu-runner binary (with file_zbc handler).
The tcmu-runner binary and corresponding libraries are pulled from
compiled source via the TCMU_RUNNER_SRC rapido.conf parameter.

On boot, the autorun script configures a backstore exposing
a local file as an emulated ZBC device via the LIO loopback fabric.

### DIFF
--- a/autorun/tcmu_zbc_loop.sh
+++ b/autorun/tcmu_zbc_loop.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+#
+# Copyright (C) SUSE LINUX GmbH 2016-2019, all rights reserved.
+#
+# This library is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published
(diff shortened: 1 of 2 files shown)


ddiss:
FWIW I haven't added CONFIG_BLK_DEV_ZONED=y to the kernel configs at this stage - that can be done later if this ends up being heavily used.

ddiss:
> Generally I like the idea, but it would be cool to have (optional) ZBC support in the blktests and fstests cut scripts?

I agree, this was mostly just meant as a starting point for more interesting tests on top. I'll close this for now and we can perhaps pick it up later if it's deemed useful. Thanks for the feedback.

PR #99, opened 2019-09-03
Bash completion fixes

ddiss:


ddiss:
> Looks good from my PoV

Thanks, merging...

PR #101, opened 2019-09-09
vm: use qemu virtio network devices

ddiss:
On my system this speeds up time to /init for a network attached VM with
static IP from 3.143863s to 0.735700s - a >4X improvement.

kernel/*config files already include CONFIG_VIRTIO_NET=y, so no other
changes should be needed.

### DIFF
--- a/vm.sh
+++ b/vm.sh
@@ -65,7 +65,7 @@ function _vm_start
 				|| _fail "IP_ADDR${vm_num} not configured"
 			kern_ip_addr="${ip_addr}:::255.255.255.0:${hostname}"
 		fi
-		qemu_netdev="-device e1000,netdev=nw1,mac=${mac_addr} \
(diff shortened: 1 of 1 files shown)


ddiss:
@morbidrsa I haven't tested your nvme RDMA over ethernet runners, but I wouldn't expect any issues there. Any thoughts? Another option would be to move the virtual network device into rapido.conf, with virtio-net as the default.

morbidrsa:
virtio-net is just another ethernet device, so it should be no problem. I'll give it a spin once I have some spare cycles just to be sure.

ddiss:
I've been running with this for well over a week now and haven't run into any issues so far. I'd like to proceed with the merge - we can always follow up with any RDMA fixes, if needed.

morbidrsa:
Agreed

PR #102, opened 2019-09-25
rapido: add fsqa user to fstests tests

morbidrsa:
Some fstests tests require the fsqa user to be present otherwise they will
be skipped.

Add everything neccessary to add the fsqa user to fstests cut scripts and
add a fsqa user and group for the fstest autorun scripts.

Signed-off-by: Johannes Thumshirn <jthumshirn@suse.de>

### DIFF
--- a/autorun/fstests_btrfs.sh
+++ b/autorun/fstests_btrfs.sh
@@ -62,6 +62,10 @@ SCRATCH_MNT=/mnt/scratch
 SCRATCH_DEV_POOL="/dev/zram1 /dev/zram2 /dev/zram3 /dev/zram4"
 EOF
 
+mkdir -p home/fsgqa
+groupadd fsgqa
+useradd -g fsgqa fsgqa
+
 set +x
 
 echo "$filesystem filesystem ready for FSQA"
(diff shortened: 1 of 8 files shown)


morbidrsa:
This can be used in ltp as well, @frankenmichl care to have a look?

ddiss:
@morbidrsa any feedback on my previous comment? My preference would be to just manually add to /etc/passwd if you think we can get away with avoiding PAM. e.g.
```
xid="2000"
ug="fsgqa"
echo "${ug}:x:${xid}:${xid}:${ug} user:/:/sbin/nologin" >> /etc/passwd
echo "${ug}:x:${xid}:" >> /etc/group
```

morbidrsa:
> @morbidrsa any feedback on my previous comment? My preference would be to just manually add to /etc/passwd if you think we can get away with avoiding PAM. e.g.
> 
> ```
> xid="2000"
> ug="fsgqa"
> echo "${ug}:x:${xid}:${xid}:${ug} user:/:/sbin/nologin" >> /etc/passwd
> echo "${ug}:x:${xid}:" >> /etc/group
> ```
Oh I totally forgot about this, but yes it should work, apart from the /sbin/nologin, which needs to be /bin/sh.

I'll re-visit it

morbidrsa:
OK, it doesn't look like **su** is able to be run without pam, but I think we can skip these tests.

ddiss:
> OK, it doesn't look like su is able to be run without pam, but I think we can skip these tests.

Hmm okay, if su can't be used without PAM then let's go ahead and add in the PAM dependencies.

PR #105, opened 2019-09-28
Adding support for native testing on non-intel archs

pevik:
Tested only on s390 and x86_64.

### DIFF
--- a/runtime.vars
+++ b/runtime.vars
@@ -1,3 +1,4 @@
+#!/bin/bash
 #
 # Copyright (C) SUSE LINUX GmbH 2016-2017, all rights reserved.
 #
@@ -11,12 +12,12 @@
 # or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public
 # License for more details.
 
-function _fail() {
+_fail() {
 	echo "error: $*"
 	exit 1
 }
 
-function _warn() {
+_warn() {
 	echo "warning: $*"
 }
 
@@ -24,7 +25,7 @@ function _warn() {
 . ${RAPIDO_DIR}/rapido.conf \
 	|| _fail "rapido.conf missing - see rapido.conf.example"
 
-function _rt_ceph_src_globals_set() {
+_rt_ceph_src_globals_set() {
 	[ -d "$CEPH_SRC" ] || _fail "$CEPH_SRC is not a directory"
 	[ -f "${CEPH_SRC}/build/CMakeCache.txt" ] \
 		|| _fail "${CEPH_SRC} cmake build missing"
@@ -43,7 +44,7 @@ function _rt_ceph_src_globals_set() {
 	CEPH_KEYRING="${CEPH_SRC}/build/keyring"
 }
 
-function _rt_ceph_installed_globals_set() {
+_rt_ceph_installed_globals_set() {
 	[ -n "$CEPH_USER" ] || _fail "CEPH_USER not configured"
 
 	# use installed Ceph binaries and configs
@@ -60,7 +61,7 @@ function _rt_ceph_installed_globals_set() {
 	CEPH_KEYRING="/etc/ceph/ceph.client.${CEPH_USER}.keyring"
 }
 
-function _rt_require_ceph {
+_rt_require_ceph() {
 	if [ -n "$CEPH_SRC" ]; then
 		_rt_ceph_src_globals_set
 	else
@@ -80,11 +81,11 @@ function _rt_require_ceph {
 	[ -f "$CEPH_KEYRING" ] || _fail "missing $CEPH_KEYRING"
 }
 
-function _ceph_get_key() {
+_ceph_get_key() {
 	${CEPH_CONF_BIN} -c ${CEPH_KEYRING} -s "client.$CEPH_USER" key 2> /dev/null
 }
 
-function _ceph_get_conf() {
+_ceph_get_conf() {
 	local section=$1
 	shift 1
 	local key="$@"
@@ -92,7 +93,7 @@ function _ceph_get_conf() {
 	${CEPH_CONF_BIN} -c ${CEPH_CONF} -s "$section" "$key" 2> /dev/null
 }
 
-function _rt_write_ceph_config() {
+_rt_write_ceph_config() {
 	local vm_ceph_conf=$1
 
 	[ -f "$vm_ceph_conf" ] || _fail "missing file $vm_ceph_conf"
@@ -136,7 +137,7 @@ function _rt_write_ceph_config() {
 # pulls in ldd dependencies, as opposed to --include "CEPH_X_BIN" "$tgt"), and
 # then use those paths directly in the autorun script (after including
 # vm_ceph.env).
-function _rt_write_ceph_bin_paths() {
+_rt_write_ceph_bin_paths() {
 	local vm_ceph_conf=$1
 
 	[ -f "$vm_ceph_conf" ] || _fail "missing file $vm_ceph_conf"
@@ -148,13 +149,12 @@ function _rt_write_ceph_bin_paths() {
 	echo "CEPH_FUSE_BIN=${CEPH_FUSE_BIN}" >> $vm_ceph_conf
 }
 
-function _rt_require_fstests() {
+_rt_require_fstests() {
 	_rt_require_conf_dir FSTESTS_SRC
 	[ -x "$FSTESTS_SRC/check" ] || _fail "missing $FSTESTS_SRC/check"
 }
 
-_rt_require_btrfs_progs()
-{
+_rt_require_btrfs_progs() {
 	local which_bin="$(which which)"
 	# sbin paths search paths may be needed for non-root
 	local p="${PATH}:/sbin:/usr/sbin"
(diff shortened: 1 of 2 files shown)


runtime.vars
```diff
@@ -153,8 +154,7 @@ function _rt_require_fstests() {
 	[ -x "$FSTESTS_SRC/check" ] || _fail "missing $FSTESTS_SRC/check"
 }
 
-_rt_require_btrfs_progs()
-{
+_rt_require_btrfs_progs() {
```
ddiss: Should probably add a "function " prefix to this one too.

runtime.vars
```diff
@@ -219,6 +219,29 @@ function _rt_require_qemu_kvm_bin {
...
+function _rt_require_arch_vars() {
+	local f="$KERNEL_SRC/.config"
+
+	KERNEL_PARAM_CONSOLE="console=ttyS0"
+	if grep -q 'CONFIG_X86.*=y' $f; then
+		ARCH_DIR="x86"
+	elif grep -q 'CONFIG_ARM=y' $f; then
+		ARCH_DIR="arm"
+	elif grep -q 'CONFIG_ARM64=y' $f; then
+		ARCH_DIR="arm64"
```
ddiss: I don't think this will work for aarch64. I had a go at this via https://github.com/ddiss/rapido/commit/cb9cc7d282d7a47d3fd25d5ef1645bccca5c9169 and it required quite a few more changes to the qemu parameters.

Please only add what you're able to test, i.e. s390 in this case.

vm.sh
```diff
@@ -16,6 +16,7 @@ RAPIDO_DIR="`dirname $0`"
 . "${RAPIDO_DIR}/runtime.vars"
 
 _rt_require_qemu_kvm_bin
+_rt_require_arch_vars
```
ddiss: I'd prefer to put the new variable exports in a `_rt_require_qemu_args` function, alongside  `QEMU_KVM_BIN` as `QEMU_KERNEL_IMG`, `QEMU_KERNEL_CONSOLE`, etc.

pevik:
Sure, I'll do both. Just a comment, I wanted to remove `function ` keyword from all (even script is for bash, this is really unnecessary bashism.

pevik:
\+ I'll try to test also `aarch64` and `ppc64le` :)

pevik:
I guess now it's better. `aarch64` and `ppc64le` aren't working for me, but maybe it's my wrong qemu configuration. Any idea what can be missing?

ddiss:
> Just a comment, I wanted to remove function keyword from all (even script is for bash, this is really unnecessary bashism.

Sounds good to me, feel free to propose a follow up for this.

runtime.vars
```diff
@@ -208,20 +208,47 @@ function _rt_require_dracut_args() {
...
+	boot_dir="$KERNEL_SRC/arch/$arch_dir/boot"
+
+	[ -d "$boot_dir" ] || \
+		_fail "directory for kernel image is missing ('$boot_dir'). Build needed?"
+
+	QEMU_KERNEL_IMG="$boot_dir/bzImage"
+	[ -f "$QEMU_KERNEL_IMG" ] || QEMU_KERNEL_IMG="$boot_dir/zImage"
+	[ -f "$QEMU_KERNEL_IMG" ] || QEMU_KERNEL_IMG="$boot_dir/uImage"
+	[ -f "$QEMU_KERNEL_IMG" ] || \
+	   _fail "no kernel image present at $QEMU_KERNEL_IMG. Wrong detection or build needed."
```
ddiss: Do we need the fallback logic here, or can we just set `QEMU_KERNEL_IMG` in/above the previous per-arch conditionals (i.e. for aarch64 set `QEMU_KERNEL_IMG=${KERNEL_SRC}/arch/arm64/boot/Image`)?

runtime.vars
```diff
@@ -208,20 +208,47 @@ function _rt_require_dracut_args() {
...
-		# Debian uses kvm instead of qemu-kvm
-		QEMU_KVM_BIN="$(which kvm 2>/dev/null)"
+	local f="$KERNEL_SRC/.config"
+	local arch_dir boot_dir qemu_bin
+
+	# TODO: add other archs from arch/*/Kconfig
+	QEMU_KERNEL_CONSOLE="console=ttyS0"
+	if grep -q 'CONFIG_X86_64=y' $f; then
+		arch_dir="x86"
+		qemu_bin="qemu-system-x86_64"
```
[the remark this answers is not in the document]
pevik: in Debian `qemu-kvm` is just shell wrapper `exec qemu-system-x86_64 -enable-kvm "$@"`, so I guess I should add `-enable-kvm` for intel. I'll double check the options for other archs.

runtime.vars
pevik: I thought that compression is configurable `CONFIG_KERNEL_{GZIP,BZIP2,LZMA,XZ}`.
But maybe compression doesn't say whether output is results to zimage, bzimage or uImage. I'll do research.

pevik:
This version work on all but powerpc (ppc64le).
There it stop producing output after:
```
Booting from memory...
Linux ppc64le
#1 SMP Sun Sep 2
```
qemu-system-ppc64 continues working, I can watch it with strace, therefore I guess it's caused by wrong `console` kernel parameter. But I tested the usual ones: console=tty0, console=ttyS0, console=ttyS0,57600, console=hvc0. Any idea on the correct one?

hramrach:
> This version work on all but powerpc (ppc64le).
> There it stop producing output after:
> 
> ```
> Booting from memory...
> Linux ppc64le
> #1 SMP Sun Sep 2
> ```
> 
> qemu-system-ppc64 continues working, I can watch it with strace, therefore I guess it's caused by wrong `console` kernel parameter. But I tested the usual ones: console=tty0, console=ttyS0, console=ttyS0,57600, console=hvc0. Any idea on the correct one?

Try watching the graphic output as well. The output may be on the graphic console that you don't see. Try -nographic to avoid that. The console attached to hvc0 is selected by firmware based on devices available (VGA, serial).

runtime.vars
hramrach: The image name is specific per-architecture with some variation for historical reasons (ie bootloader searching for a specific filename). The filename is not related to content.

ie ppc64 uses the plain unwrapped uncompressed ELF image which is normally installed as boot/vmlinux but it is sometimes installed as /boot/vmlinuz because historical bootloader scripts would search for this filename and the bootloader can detect the image format correctly (or supports only one format anyway).

pevik:
Current code fails on `invalid magic at start of compressed archive`:

`CONFIG_BLK_DEV_RAM=m`
```
[    0.143098] Unpacking initramfs...
[    0.143136] Initramfs unpacking failed: invalid magic at start of compressed archive
```
or `CONFIG_BLK_DEV_RAM is not set`
```
[    0.145764] Trying to unpack rootfs image as initramfs...
[    0.145866] rootfs image is not initramfs (invalid magic at start of compressed archive); looks like an initrd
```
Again, `qemu-system-ppc64` is not crashed (watched with `strace`), but no output is produced.
Any idea, what can be wrong?

pevik:
Hm, testing https://github.com/linuxppc/wiki/wiki/Booting-with-Qemu

```
wget https://openpower.xyz/job/initramfs/job/buildroot-master/lastSuccessfulBuild/artifact/rootfs-le.cpio.xz
qemu-system-ppc64 -M pseries -m 1G -nographic -vga none -kernel zImage -initrd rootfs-le.cpio.xz
```
And it fails the same. Do I miss some compile option (note: using `zImage`, not `vmlinux`).

hramrach:
It works with SLE ramdisk and SLE kernel so you probably build the ramdisk incorrectly. Note the ramdisk name specifies LE so it may be incompatible between BE and LE. Also not compiling in support for the compression used can cause this issue.

hramrach:
> Current code fails on `invalid magic at start of compressed archive`:
> 
> kernel compiled with `CONFIG_BLK_DEV_RAM=m`

If you want to use this option it should be set to y

The modules are in the ramdisk and if ramdisk support is in a module there is no way to get at it.

pevik:
Now it's ready, all 4 archs are working.

`ppc64le` required using `vmlinux`. I also have entropy issues on s390 (`random: lvm: uninitialized urandom read (4 bytes read)`, while `haveged` is running), but that's another story).

ddiss:
> Now it's ready, all 4 archs are working.

Thanks! I'll need a bit of time to test this.

> ppc64le required using vmlinux. I also have entropy issues on s390 (random: lvm: uninitialized urandom read (4 bytes read), while haveged is running), but that's another story).

virtio-rng is used to work around entropy issues on x86-64. rapido.conf.example specifies:
 QEMU_EXTRA_ARGS="-nographic -device virtio-rng-pci"

pevik:
Sure, take your time.
Ad randomness: `-device virtio-rng-pci` is not supported on s390 (qemu quits), using `-device virtio-rng-ccw` does not help. It's strange, that previously s390 worked, entropy problems started after machine reinstallation, so it's probably a configuration issue. I haven't experienced entropy problems on other archs.

ddiss:
Merged - thanks a lot for your work here.

Issue #106, opened 2019-10-25
Common function to create /etc/hosts file


luis-henrix:
Several autorun scripts were creating /etc/hosts.  Add a common function for doing that.  While there, create this file following the recommendations from hosts(5), namely:
    
- use 127.0.0.1 for localhost
- use 127.0.1.1 for the FQDN
- add IPv6 entries

ddiss:
Merged, thanks.

PR #107, opened 2019-10-25
vm_autorun.env: add function to create /etc/hosts file #106

luis-henrix:
This PR fixes issue #106 . Here's the commit changelog:

Several autorun scripts were creating /etc/hosts.  Add a common function
for doing that.  While there, create this file following the
recommendations from hosts(5), namely:

 - use 127.0.0.1 for localhost
 - use 127.0.1.1 for the FQDN
 - add IPv6 entries

Signed-off-by: Luis Henriques <lhenriques@suse.com>

### DIFF
--- a/autorun/fstests_btrfs.sh
+++ b/autorun/fstests_btrfs.sh
@@ -23,14 +23,9 @@ set -x
 
 [ -n "$BTRFS_PROGS_SRC" ] && export PATH="${PATH}:${BTRFS_PROGS_SRC}"
 
-hostname_fqn="`cat /proc/sys/kernel/hostname`" || _fatal "hostname unavailable"
-hostname_short="${hostname_fqn%%.*}"
-filesystem="btrfs"
+_vm_ar_hosts_create
 
-# need hosts file for hostname -s
-cat > /etc/hosts <<EOF
-127.0.0.1	$hostname_fqn	$hostname_short
-EOF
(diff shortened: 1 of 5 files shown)


luis-henrix:
> My machine carries "ipv6-" instead of "ip6-" prefixes for the special ipv6 addresses, but I noticed that https://bugs.launchpad.net/ubuntu/+source/netbase/+bug/301430 includes the latter. Any ideas what's correct here?

TBH I don't really know which one is correct and I couldn't find any reference about this.  I simply used what's recommend in hosts(5) man page.  But my TW workstation also shows the "ipv6-" prefix, so I'm really not sure which one to pick...

> Also one minor nit: please declare the "hostname_" variables as local.

Done.

ddiss:
> > My machine carries "ipv6-" instead of "ip6-" prefixes for the special ipv6 addresses, but I noticed that https://bugs.launchpad.net/ubuntu/+source/netbase/+bug/301430 includes the latter. Any ideas what's correct here?
> 
> TBH I don't really know which one is correct and I couldn't find any reference about this. I simply used what's recommend in hosts(5) man page. But my TW workstation also shows the "ipv6-" prefix, so I'm really not sure which one to pick...

Thanks for investigating. I'll merge what's here for now and we can follow up with this later (which may include dropping the ip6 aliases).

PR #109, opened 2019-11-04
vm: support nested virtualization

ddiss:
Only attempt to use QEMU KVM acceleration if /dev/kvm is present.


morbidrsa:
Merged

PR #110, opened 2019-11-05
kernel: add sle15sp2 config

ddiss:
Based on sle15sp1_config. xxhash and zstd support enabled, otherwise
nothing too interesting.


ddiss:
Ping - any takers?

PR #114, opened 2019-12-08
add self test coverage

ddiss:
selftest/selftest.sh runs a quick check to ensure that rapido can cut and boot the simple-example image.

### DIFF
--- a/rapido.conf.example
+++ b/rapido.conf.example
@@ -47,6 +47,10 @@ TAP_USER=""
 # e.g. QEMU_EXTRA_ARGS="-nographic -device virtio-rng-pci -gdb tcp:127.0.0.1:1234"
 QEMU_EXTRA_ARGS="-nographic -device virtio-rng-pci"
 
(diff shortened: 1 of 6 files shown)


rapido.conf.example
```diff
@@ -47,6 +47,10 @@ TAP_USER=""
 # e.g. QEMU_EXTRA_ARGS="-nographic -device virtio-rng-pci -gdb tcp:127.0.0.1:1234"
 QEMU_EXTRA_ARGS="-nographic -device virtio-rng-pci"
 
+# extra kernel boot parameters, passed via QEMU -append
+# e.g. QEMU_EXTRA_KERNEL_PARAMS="loglevel=0"
+#QEMU_EXTRA_KERNEL_PARAMS=""
```
morbidrsa: Coincidently I have the same change in my local version, seems like QEMU_EXTRA_KERNEL_PARAMS is really needed

PR #122, opened 2020-02-07
cut/cephfs: drop CEPH_MOUNT_BIN helper binary

ddiss:
It's not needed, and has a bunch of dependencies. ceph.conf and the
keyring can also be removed thanks to _rt_write_ceph_config().


ddiss:
> This looks OK to me. I've used mount.ceph before for testing the tool itself, but to be honest I can't remember if I really used it from within the VM or on the host itself :-) Either way, I don't see a problem with dropping it.

Thanks. Let's revisit this if we end up wanting to have test coverage for mount.ceph, but for now I'd like to drop it, just to clean up dependencies.

PR #123, opened 2020-02-10
More minor btrfs fstests updates

morbidrsa:


ddiss:
> The changes look fine, but please add xfs_freeze and fsck to cut/fstests_xfs.sh too.

actually xfs_freeze is already there, so just fsck

morbidrsa:
added fsck to xfs as well

PR #125, opened 2020-03-20
Zonefstests

morbidrsa:
Add zonefs testing to rapido.


cut/zonefstests_nullblk.sh
```diff
@@ -0,0 +1,35 @@
...
+
+RAPIDO_DIR="$(realpath -e ${0%/*})/.."
+. "${RAPIDO_DIR}/runtime.vars"
+
+_rt_require_dracut_args
+
+"$DRACUT" \
+	--install "ps rmdir dd id basename stat wc grep blkzone cut fio \
+		   rm truncate" \
+	--include "$ZONEFSTOOLS_SRC/src/mkzonefs" "/usr/bin/mkzonefs" \
```
ddiss: does `mkzonefs` have any shared object dependencies? If so, I'd prefer to add it to the `--install` list and add `$ZONEFSTOOLS_SRC/src` to `PATH` in the autorun file, like we do for fstests.

cut/zonefstests_nullblk.sh
```diff
@@ -0,0 +1,35 @@
...
+#
+# This library is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public
+# License for more details.
+
+RAPIDO_DIR="$(realpath -e ${0%/*})/.."
+. "${RAPIDO_DIR}/runtime.vars"
+
+_rt_require_dracut_args
```
ddiss: Please add `_rt_require_conf_dir ZONEFSTOOLS_SRC` here

cut/zonefstests_nullblk.sh
morbidrsa: It actually does have libuiid.so and libblkid.so, but I don't think mkzonefs is packaged for any distribution yet.

I'll look at how we're resolving this in fstests.

cut/zonefstests_nullblk.sh
morbidrsa: done

cut/zonefstests_nullblk.sh
morbidrsa: done

ddiss:
Merged - thanks!

PR #128, opened 2020-05-29
tools/zram_hot_add: only add a single device

ddiss:
If called before the zram module is loaded, two zram devices appear.
This is due to the default num_devices=1 probe behaviour and
unconditional sysfs zram-control/hot_add use.


ddiss:
Ping, any objections to this?

morbidrsa:
Oops,  totally forgot about this one, looks good.

PR #131, opened 2020-06-24
rapdio: add xxhash_generic module to btrfs initrd

morbidrsa:
Since kernel v5.5 btrfs can use xxhash64 and sha256 as checksum hashes,
but we don't add the xxhash_generic driver to rapdio's initrd.

Add xxhash_generic to initrd, sha256 isn't needed as the rapido kernel
config has sha256 built in.

Signed-off-by: Johannes Thumshirn <johannes.thumshirn@wdc.com>


morbidrsa:
Do other btrfs consumers (e.g. samba-local) require this module?

No, normally btrfs is using crc32c as checksum algorithm so this is fine. It's only if the FS is formatted with xxhash64 as checksum algorithm that the module is actually needed by the kernel, otherwise a mount will fail.

PR #133, opened 2020-07-22
runtime.vars: drop dracut --list-modules sanity check

ddiss:
On my system this costs ~200ms every "cut" invocation. It's unnecessary,
as dracut invocation failures will be caught when we actually generate
the initramfs image.


morbidrsa:
Looks good

PR #135, opened 2020-08-07
rapdio: add environment for zonefs tests over SCSI

morbidrsa:
Now that the kernel' scsi_debug driver can emulate zoned block devices, we
can add environment for zonefs tests over SCSI.

This way we also have regression tests for SCSI only callpaths zonefs will
take, like the emulation of the ZONE APPEND commands in the SCSI layer.

Signed-off-by: Johannes Thumshirn <johannes.thumshirn@wdc.com>

### DIFF
--- a/autorun/zonefstests_scsi.sh
+++ b/autorun/zonefstests_scsi.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+#
+# Copyright (C) 2020 Western Digital Corporation, all rights reserved.
+#
+# This library is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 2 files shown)


morbidrsa:
There are no newly introduced config options. It is basically the zonefstests-nullblk scripts just with scsi_debug instead of null_blk.

ddiss:
> There are no newly introduced config options. It is basically the zonefstests-nullblk scripts just with scsi_debug instead of null_blk.

Sorry, I overlooked that they're already carried in conf.example. Thanks, merging...

PR #136, opened 2020-08-26
kernel/vanilla_config: update to 5.9.0-rc2

ddiss:
Nothing new enabled. We might want to evaluate moving from KERNEL_GZIP
to something with better decompression speed in future.


ddiss:
The follow-up CPUFreq change basically halves VM boot time on my laptop. Given the time saving I'll probably also enable it for some of the older SLE configs.

ddiss:
> The follow-up CPUFreq change basically halves VM boot time on my laptop. Given the time saving I'll probably also enable it for some of the older SLE configs.

Hmm actually the speedup doesn't seem to be reproducible (this time with a 15sp2 kernel). I'm guessing the boot time fluctuations might be due to the dynamic powersave governor on my laptop getting in the way. Either way, enabling CPUFreq perf doesn't appear to cost us anything, so I think it's still worth it.

ddiss:
> Hmm actually the speedup doesn't seem to be reproducible (this time with a 15sp2 kernel). I'm guessing the boot time fluctuations might be due to the dynamic powersave governor on my laptop getting in the way. Either way, enabling CPUFreq perf doesn't appear to cost us anything, so I think it's still worth it.

I did some more testing with mainline, observing the same fluctuations. I'm going to withdraw the cpufreq change until we actually need it.

morbidrsa (approved):
Can't really comment on SLE but vanilla looks good to me

ddiss:
> Can't really comment on SLE but vanilla looks good to me

Thanks Johannes, merging...

PR #137, opened 2020-09-04
Use dracut emergency hook

ddiss:
This patchset fixes systemd based boots and also significantly cleans up the way autorun scripts are invoked.
Instead of hoping that the dracut emergency shell (after dumping rdsosreport) will magically trigger `/.profile`, we now use a proper emergency hook which is invoked earlier and allows complete control over subsequent autorun invocation.

### DIFF
--- a/autorun/00-rapido-init.sh
+++ b/autorun/00-rapido-init.sh
@@ -14,8 +14,6 @@
 # This script runs as the first Dracut entry point during /init.
 # $DRACUT_SYSTEMD is set when run via systemd.
 
-echo "Rapido: starting autorun script..."
-
 _ctty=
 for i in $(cat /proc/cmdline); do
 	case "$i" in
--- a/autorun/00-rapido-init.sh
+++ b/autorun/00-rapido-init.sh
@@ -1,0 +1,33 @@
+#
+# Copyright (C) SUSE LLC 2020, all rights reserved.
+#
+# This library is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published
+# by the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) version 3.
+#
+# This library is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public
+# License for more details.
+
+# This script runs as the first Dracut entry point during /init.
+# $DRACUT_SYSTEMD is set when run via systemd.
+
+echo "Rapido: starting autorun script..."
+
+_ctty=
+for i in $(cat /proc/cmdline); do
+	case "$i" in
+		"console="*)
+			_ctty="/dev/${i#console=}"
+			break
+			;;
+	esac
+done
+[ -c "$_ctty" ] || _ctty=/dev/tty1
+setsid --ctty /bin/sh -i -l 0<>$_ctty 1<>$_ctty 2<>$_ctty
+
+# shut down when rapido autorun / shell exits...
+echo 1 > /proc/sys/kernel/sysrq && echo o > /proc/sysrq-trigger
+sleep 20
--- a/cut/blktests_rbd.sh
+++ b/cut/blktests_rbd.sh
@@ -19,7 +19,7 @@ vm_ceph_conf="$(mktemp --tmpdir vm_ceph_conf.XXXXX)"
 # remove tmp file once we're done
 trap "rm $vm_ceph_conf" 0 1 2 3 15
 
-_rt_require_dracut_args
+_rt_require_dracut_args "$RAPIDO_DIR/autorun/blktests_rbd.sh"
 _rt_require_ceph
 _rt_write_ceph_config $vm_ceph_conf
 _rt_require_blktests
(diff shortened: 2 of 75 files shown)


autorun/00-rapido-init.sh
```diff
@@ -0,0 +1,30 @@
...
+echo "Rapido: starting autorun script..."
+
+_ctty=console
+while [ -f /sys/class/tty/$_ctty/active ]; do
+	_ctty=$(cat /sys/class/tty/$_ctty/active)
+	_ctty=${_ctty##* } # last one in the list
+done
+_ctty=/dev/$_ctty
+[ -c "$_ctty" ] || _ctty=/dev/tty1
+setsid --ctty /bin/sh -i -l 0<>$_ctty 1<>$_ctty 2<>$_ctty
```
ddiss, line 26: I should mention that the above `_ctty` magic above is mostly taken from Dracut... I think we should be able to just hardcode the device path based on the `console=X` boot parameter.

luis-henrix: Yeah, it's a bit odd but it seems to be working fine (although I can't really say I understand the rational behind this magic...)

cut/blktests_rbd.sh
```diff
@@ -36,6 +36,8 @@ _rt_require_blktests
 	--include "$RBD_UDEV_RULES" "/usr/lib/udev/rules.d/50-rbd.rules" \
 	--include "$BLKTESTS_SRC" "/blktests" \
 	--include "$RAPIDO_DIR/autorun/blktests_rbd.sh" "/.profile" \
+	--include "$RAPIDO_DIR/autorun/00-rapido-init.sh" \
```
luis-henrix, line 39: This is just a suggestion, but maybe it's worth setting up a new variable to runtime.vars that would have all the '--include...' statements that are required in all (or most) of the 'cut' scripts.  This would make future maintenance a bit easier.

luis-henrix (approved):
I've only tested a few of the 'cut' scripts, but these changes look good to me.  The only suggestion I had was the new variable to the 'cut' scripts, but... that's just a suggestion ;-)

cut/blktests_rbd.sh
ddiss: I agree, I'll rework this to use a `DRACUT_RAPIDO_INCLUDES` (or similar) global for this. Thanks for the review :)

ddiss: Hmm, I've done this, but don't much like how it conceals the presence of /rapido.conf and /vm_autorun.env to people writing autorun scripts. I'll have a bit more of a think about this and drop the `DRACUT_RAPIDO_INCLUDES` change for now.

ddiss: Okay, I've worked on this further and have something which I'm much happier with. The autorun script is provided as a param for `_rt_require_dracut_args()`, which handles putting it in the right place for the init sequence.
autorun scripts are sourced directly from `vm_autorun.env` (which now lives at `.profile`), so no longer need to source it explicitly.

autorun/00-rapido-init.sh
ddiss: I've got a patch which changes this to use the vm.sh console parameter and will push it with v2.

ddiss:
changes since v1:
* use kernel console= parameter to derive ctty setting
* use new DRACUT_RAPIDO_INCLUDES variable for rapido core init and autorun includes parameters
* rework boot chain to source autorun script from within vm_autorun.env
  * drop explicit vm_autorun.env includes from autorun scripts
* slightly rework autorun-is-running-from-within-rapido-vm safety check

@luis-henrix it'd be great if you could also take a look at this new patchset.

ddiss:
changes since v2:
- support specifying and running multiple autorun scripts
- use this functionality for /vm_ceph.env content and drop explicit include

I'm really starting to like this series. It makes the dracut -> autorun boot process a lot more predictable and user friendly without hiding away too much behind layers of abstraction. Better still, it removes a heap of boilerplate code from autorun scripts **and** improves boot times.

runtime.vars
```diff
@@ -177,9 +177,12 @@ _rt_require_blktests() {
 }
 
 _rt_require_dracut_args() {
-	_rt_require_conf_dir KERNEL_SRC
-	local kver="$(cat ${KERNEL_SRC}/include/config/kernel.release)"
-	[ -n "$kver" ] || _fail "failed to read kernel.release at $KERNEL_SRC"
+	# first arg must correspond to a script that will be run on boot
+	local rapido_init=$1
+	[ -n "$rapido_init" ] \
+		|| _fail "_rt_require_dracut_args missing autorun script param"
```
luis-henrix, line 186: Minor comment (feel free to ignore): Maybe instead of failing the function could try to check if there was an 'autorun' script with the same name as the 'cut' script.  This would make the parameter unneeded for most (all?) cases.  Although maybe a bit more obscure and less obvious for anyone writing a new 'cut' script... In case you decide to implement this suggestion, maybe it's easier to merge this change with the commit that adds support for multiple scripts (which is also a **really** cool idea!)

luis-henrix (approved):
I think this series is really awesome.  It simplifies a lot of the scripts.  Feel free to ignore the only comment I had; I'm not even sure if what I mention in that comment is really a good idea ;-)

runtime.vars
ddiss: I think I'd prefer to keep autorun scripts explicitly named in the cut scripts for now. cut scripts could still autogenerate the autorun script path using `basename` or something.

Thanks a lot for your review, Luis! I'll go ahead and do the merge...

Issue #89, opened 2019-06-22
move VM network configuration into separate config file?


ddiss:
(brainstorming ways to add support for >2 network attached VMs)

Including VM network configuration in rapido.conf means that users need to learn yet another network configuration syntax. It also adds clutter to rapido.conf, particularly if we wish to support deployment of more than two VMs in future (https://github.com/rapido-linux/rapido/pull/51 added three, but it turned into a bit of a mess).

It should be possible to have br_setup.sh run dnsmasq with a pregenerated config file that offers sensible defaults for all VMs (tap MAC addresses are used for corresponding VM nics). Advanced users could provide their own settings or network environment, allowing for more flexibility while keeping the complexity out of rapido.

As a trade off, such an architecture would perhaps mean dropping support for static VM IP assignment.

ddiss:
cc'ing @scabrero  here...

Considering this a little more, I think that `sysconfig/network` syntax might also work. E.g. we have:
`rapido-net-conf/ifcfg-br#` <- bridge configuration, one per desired bridge interface
`rapido-net-conf/vmX/` <- directory per VM. rapido's VM limit could then also be extended (in future) based on how many dirs exist
`rapido-net-conf/vmX/ifcfg-Y` <- ifcfg syle config

Alternatively, maybe there's a container engine json network config syntax that would also suite our needs?

ddiss:
More research questions:
- does `wicked` require `dbus` at runtime?
  + is it only available on SUSE based distros?
- `systemd-network` (https://www.freedesktop.org/software/systemd/man/systemd.network.html ) looks promising and quite lightweight. Although that may mean that all other systemd dependencies also need to be dragged in?

ddiss:
Fixed via https://github.com/rapido-linux/rapido/pull/173

PR #143, opened 2020-10-19
Add a second network interface to VMs

scabrero:
Changes:

* Rename configuration variables to use `BRx_` and `VMx_` as prefix
* Allow to skip the bridge device provision with `BRx_DEV_SKIP_PROVISION`, useful to reuse the bridge devices created by libvirt and connect the rapido VMs to them
* Allow to create a second network adapter in each VM, connected to the second bridge device `BR2_DEV`.
* Setup the network in `vm_autorun.env` instead of using the `ip=` kernel parameter. It will allow to use IPv6 addresses in the future.

### DIFF
--- a/tools/br_setup.sh
+++ b/tools/br_setup.sh
@@ -15,7 +15,7 @@
 RAPIDO_DIR="$(realpath -e ${0%/*})/.."
 . "${RAPIDO_DIR}/runtime.vars"
 
-_rt_require_conf_setting BR1_DEV TAP_USER TAP_DEV0 TAP_DEV1
+_rt_require_conf_setting BR1_DEV TAP_USER VM1_TAP_DEV1 VM2_TAP_DEV1
 
 # cleanup on premature exit by executing whatever has been prepended to @unwind
 unwind=""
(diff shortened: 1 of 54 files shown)


tools/br_setup.sh
```diff
@@ -15,63 +15,145 @@
...
+		echo -n " with address $BR1_ADDR"
+	fi
+
+	if [ -n "$BR1_IF" ]; then
+		ip link set $BR1_IF master $BR1_DEV || exit 1
+		unwind="ip link set $BR1_IF nomaster; ${unwind}"
+		echo -n ", connected to $BR1_IF"
+	fi
+	echo
 fi
```
ddiss, line 41: Instead of adding the `BR1_DEV_SKIP_PROVISON` param, could we just ignore any preexisting bridge interface here?

ddiss (changes requested):
Thanks a lot for your changes here, Samuel.

Aside from the minor inline comments I have two concerns regarding the approach:
- although very helpful for some use-cases, I'm worried about the configuration complexity added for new users here.
  + perhaps we could reuse some other existing config syntax (sysconfig ifcfg or dnsmasq) so that users don't need to learn something new, and we don't need to continuously to add new rapido.conf options
  + I raised this approach a while back via https://github.com/rapido-linux/rapido/issues/89
- this will break users' with existing rapido.conf network configs, requiring them to rewrite them.
  + I don't want to support old and new syntax, but maybe some sort of upgrade helper would be worth while?
- moving from kernel dhcp client to `dhclient` seem like a good idea (finally we can support standard distro kernels), but it does appear to add a little bit of bloat to images

ddiss:
> Thanks a lot for your changes here, Samuel.
> 
> Aside from the minor inline comments I have some concerns regarding the approach:
> 
>     * although very helpful for some use-cases, I'm worried about the configuration complexity added for new users here.
>       
>       * perhaps we could reuse some other existing config syntax (sysconfig ifcfg or dnsmasq) so that users don't need to learn something new, and we don't need to continuously to add new rapido.conf options
>       * I raised this approach a while back via #89
> 
>     * this will break users' with existing rapido.conf network configs, requiring them to rewrite them.
>       
>       * I don't want to support old and new syntax, but maybe some sort of upgrade helper would be worth while?
> 
>     * moving from kernel dhcp client to `dhclient` seem like a good idea (finally we can support standard distro kernels), but it does appear to add a little bit of bloat to images

@scabrero any thoughts on how you think we should proceed here?

tools/br_setup.sh
scabrero: This is a good idea. I will change it.

scabrero:
> > Thanks a lot for your changes here, Samuel.
> > Aside from the minor inline comments I have some concerns regarding the approach:
> > ```
> > * although very helpful for some use-cases, I'm worried about the configuration complexity added for new users here.
> >   
> >   * perhaps we could reuse some other existing config syntax (sysconfig ifcfg or dnsmasq) so that users don't need to learn something new, and we don't need to continuously to add new rapido.conf options
> >   * I raised this approach a while back via #89
> > 
> > * this will break users' with existing rapido.conf network configs, requiring them to rewrite them.
> >   
> >   * I don't want to support old and new syntax, but maybe some sort of upgrade helper would be worth while?
> > 
> > * moving from kernel dhcp client to `dhclient` seem like a good idea (finally we can support standard distro kernels), but it does appear to add a little bit of bloat to images
> > ```
> 
> @scabrero any thoughts on how you think we should proceed here?

I had a look to #89 and I am wondering, what do you think if we add systemd and systemd-networkd to all images and provide the systemd-networkd files for each vm in `rapido-net-conf/vmX/`, injecting them into the initramfs image using DRACUT_RAPIDO_INCLUDES?

We could extend it and install systemd-resolved too to have DNS resolution in the VMs.

ddiss:
> I had a look to #89 and I am wondering, what do you think if we add systemd and systemd-networkd to all images and provide the systemd-networkd files for each vm in rapido-net-conf/vmX/, injecting them into the initramfs image using DRACUT_RAPIDO_INCLUDES?

I'd like to keep dependencies/images minimal, so using systemd-networkd only would be my preference if possible (it's under 500k and appears to be quite powerful). I had a play with it myself, but couldn't get it to bring up a dhcp client address by itself. In the end all I really want is an clean network config syntax which could be used for configuring an arbitrary number of NICs per-VM, with static or dhcp based v4 or v6 addresses, etc., preferably outside of rapido.conf .

ddiss:
> I'd like to keep dependencies/images minimal, so using systemd-networkd only would be my preference if possible (it's under 500k and appears to be quite powerful). I had a play with it myself, but couldn't get it to bring up a dhcp client address by itself. In the end all I really want is an clean network config syntax which could be used for configuring an arbitrary number of NICs per-VM, with static or dhcp based v4 or v6 addresses, etc., preferably outside of rapido.conf .

Just an update here - I plan on retesting with an updated systemd-networkd once I upgrade to openSUSE 15.3. @scabrero given that you had systemd-networkd working standalone on Tumbleweed, I'm hoping it's just the newer version that's needed before we can merge here.

ddiss:
> Just an update here - I plan on retesting with an updated systemd-networkd once I upgrade to openSUSE 15.3. @scabrero given that you had systemd-networkd working standalone on Tumbleweed, I'm hoping it's just the newer version that's needed before we can merge here.

The systemd-networkd approach tracked via https://github.com/rapido-linux/rapido/pull/147 seems to be progressing quite well, and should offer a cleaner and more flexible configuration format supporting multihoming. @scabrero I'm going to close this, but feel free to reopen if you feel strongly about keeping these changes.

PR #145, opened 2020-11-12
tcmu_file_iscsi: cut+autorun scripts for tcmu-runner iscsi

ddiss:
Use the simple tcmu-runner file backend to expose a local file via
iSCSI.

### DIFF
--- a/autorun/tcmu_file_iscsi.sh
+++ b/autorun/tcmu_file_iscsi.sh
@@ -1,0 +1,88 @@
+#!/bin/bash
+#
+# Copyright (C) SUSE LLC 2020, all rights reserved.
+#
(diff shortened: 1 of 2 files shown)


ddiss:
Changes since v1:
- give tcmu-runner time to start before configuration
- truncate image file to desired size

@morbidrsa / @luis-henrix : mind if I merge this?

luis-henrix:
I've reviewed this commit and the only thing that caught my attention was how small the cut script was :-)
Unfortunately, I wasn't able to test it yet.  I'm OK with this being merged as-is.  However, if you'd rather have some more testing done, I can plan for doing so during next week.

ddiss:
Thanks Luis :-)
I've tested it quite a bit locally, so will merge as-is.

PR #146, opened 2020-11-27
support post-autorun scripts via rapido cut -f <script> and -x <command> parameters

ddiss:
We currently provide rapido.conf `FSTESTS_AUTORUN_CMD`, `BLKTESTS_AUTORUN_CMD`, etc. parameters for running arbitrary commands after the autorun script has setup the test environment.
This patchset adds new rapido cut -f <script> and -x <command> parameters to run an arbitrary script file or command after autorun has completed. Multiple <script> or <command> parameters can be specified, and they will be executed in the order of appearance. E.g.
```
./rapido cut -x "echo run me..." -x "echo after autorun" simple-example
...
Rapido: starting /rapido_autorun/00-simple_example.sh
...
Rapido scratch VM running. Have a lot of fun...
Rapido: starting /rapido_autorun/01-rapido_post_autorun.6q9lTJL9EU
run me...
Rapido: starting /rapido_autorun/02-rapido_post_autorun.k2pWEUYvwy
after autorun
rapido1:/#
```
IMO this offers a more flexible approach than `FSTESTS_AUTORUN_CMD`, etc. making integration with kernel CI utilities much more straightforward.

@morbidrsa / @luis-henrix : interested to hear your thoughts on this.

### DIFF
--- a/cut/blktests_rbd.sh
+++ b/cut/blktests_rbd.sh
@@ -19,7 +19,8 @@ vm_ceph_conf="$(mktemp --tmpdir vm_ceph_conf.XXXXX)"
 # remove tmp file once we're done
 trap "rm $vm_ceph_conf" 0 1 2 3 15
 
-_rt_require_dracut_args "$vm_ceph_conf" "$RAPIDO_DIR/autorun/blktests_rbd.sh"
+_rt_require_dracut_args "$vm_ceph_conf" "$RAPIDO_DIR/autorun/blktests_rbd.sh" \
+			"$@"
 _rt_require_ceph
 _rt_write_ceph_config $vm_ceph_conf
 _rt_require_blktests
--- a/cut/blktests_zram.sh
+++ b/cut/blktests_zram.sh
@@ -15,7 +15,7 @@
 RAPIDO_DIR="$(realpath -e ${0%/*})/.."
 . "${RAPIDO_DIR}/runtime.vars"
 
-_rt_require_dracut_args "$RAPIDO_DIR/autorun/blktests_zram.sh"
+_rt_require_dracut_args "$RAPIDO_DIR/autorun/blktests_zram.sh" "$@"
 _rt_require_blktests
 
 "$DRACUT" --install "tail blockdev ps rmdir resize dd vim grep find df sha256sum \
--- a/cut/cephfs.sh
+++ b/cut/cephfs.sh
@@ -21,7 +21,7 @@ trap "rm $vm_ceph_conf" 0 1 2 3 15
 
 _rt_require_ceph
 _rt_write_ceph_config $vm_ceph_conf
-_rt_require_dracut_args "$vm_ceph_conf" "$RAPIDO_DIR/autorun/cephfs.sh"
+_rt_require_dracut_args "$vm_ceph_conf" "$RAPIDO_DIR/autorun/cephfs.sh" "$@"
 
 "$DRACUT" --install "tail blockdev ps rmdir resize dd vim grep find df sha256sum \
 		   strace stat which touch cut chmod true false \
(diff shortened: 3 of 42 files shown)


morbidrsa:
I like the overall idea very much. I tend to also use rapido with syzbot reproducers, and this makes it way easier to just run a custom script.

Will look at the code soon as well.

morbidrsa (commented):
I can't find where the file passed as argument of -f is actually added to the image. All I see here is that it's passed to the $cut_script, but none of the cut scripts actually do anything with it.

morbidrsa (commented):
Where does it get installed and executed?

I've pulled your branch and tried it using the following commands:
> ./rapido cut fstests-btrfs -f syz-repro -x syz-repro

but in the installed image I can't find the passed sys-repro, nor does it even try to execute it.

ddiss:
> I can't find where the file passed as argument of -f is actually added to the image. All I see here is that it's passed to the $cut_script, but none of the cut scripts actually do anything with it.

The cut scripts pass all remaining arguments through to `_rt_require_dracut_args`, which adds the files to the `/rapido_autorun/` directory for execution on boot.

> I've pulled your branch and tried it using the following commands:

>    ./rapido cut fstests-btrfs -f syz-repro -x syz-repro

The new parameters need to go directly after `cut` and before the cut script name.

morbidrsa:
> > I can't find where the file passed as argument of -f is actually added to the image. All I see here is that it's passed to the $cut_script, but none of the cut scripts actually do anything with it.
> 
> The cut scripts pass all remaining arguments through to `_rt_require_dracut_args`, which adds the files to the `/rapido_autorun/` directory for execution on boot.
> 
> > I've pulled your branch and tried it using the following commands:
> 
> > ./rapido cut fstests-btrfs -f syz-repro -x syz-repro
> 
> The new parameters need to go directly after `cut` and before the cut script name.

Ah ok, got it working now. Although it's a bit cumbersome to just add a binary and a shellscript and then run the script calling the binary.

ddiss:
> Ah ok, got it working now. Although it's a bit cumbersome to just add a binary and a shellscript and then run the script calling the binary.

Yeah, it can only be used to add a boot script (without any new dependencies). We could add extra parameters for dracut `--include` / `--install`, but I don't really like the direction we're heading... How about adding a new cut / autorun testcase which adds all of the syzkaller dependencies?

morbidrsa:
> > Ah ok, got it working now. Although it's a bit cumbersome to just add a binary and a shellscript and then run the script calling the binary.
> 
> Yeah, it can only be used to add a boot script (without any new dependencies). We could add extra parameters for dracut `--include` / `--install`, but I don't really like the direction we're heading... How about adding a new cut / autorun testcase which adds all of the syzkaller dependencies?

Nah no need, I'll usually just hack something temporarily into the cut scripts.

ddiss:
> Nah no need, I'll usually just hack something temporarily into the cut scripts.

Same :). Which begs the question whether this patchset would still be useful... For me I think it'd still be helpful for CI type stuff, where some hooks for test result parsing / shutdown, etc. may be helpful.

morbidrsa:
> > Nah no need, I'll usually just hack something temporarily into the cut scripts.
> 
> Same :). Which begs the question whether this patchset would still be useful... For me I think it'd still be helpful for CI type stuff, where some hooks for test result parsing / shutdown, etc. may be helpful.

Yep I think it's very useful.

ddiss:
@morbidrsa I've added test coverage for the new options, and fixed handling so that unhandled parameters provided after the cut image name (e.g. `./rapido cut fstests-btrfs -f syz-repro.sh`) generate an error. Please let me know if you're okay with it.

morbidrsa:
I won't have a chance to look deeply into it this year. But from a high level PoV it looks good. I think we can merge it.

PR #147, opened 2020-12-15
WIP Use systemd-networkd to configure the network

scabrero:
Remaining issues and possible solutions:

- Drop IP_ADDR1 and IP_ADDR2 references in autorun scripts
  Get the address from the interface at run time

- Fix br_setup static DHCP leases
  Redefine the IP_ADDR1_DHCP variable to be the static lease address instead a boolean.

- Remove IP_ADDR1, IP_ADDR2, IP_ADDR1_DHCP and IP_ADDR2_DHCP variables

### DIFF
--- a/cut/mpath_local.sh
+++ b/cut/mpath_local.sh
@@ -28,9 +28,10 @@ _rt_require_dracut_args "$RAPIDO_DIR/autorun/mpath_local.sh" "$@"
 # Once booted, you can simulate path failure by switching to the QEMU console
 # (ctrl-a c) and running "drive_del hda"
 
-"$DRACUT" --install "tail blockdev ps rmdir resize dd vim grep find df sha256sum \
-		   strace mkfs mkfs.xfs parted partprobe sgdisk hdparm \
-		   timeout id chown chmod env killall getopt basename" \
+"$DRACUT" --install "$DRACUT_RAPIDO_INSTALLS \
+		tail blockdev ps rmdir resize dd vim grep find df sha256sum \
+		strace mkfs mkfs.xfs parted partprobe sgdisk hdparm \
+		timeout id chown chmod env killall getopt basename" \
 	$DRACUT_RAPIDO_INCLUDES \
 	--add-drivers "virtio_scsi virtio_pci sd_mod" \
 	--modules "$DRACUT_RAPIDO_MODULES systemd systemd-initrd dracut-systemd multipath" \
--- a/cut/mpath_local.sh
+++ b/cut/mpath_local.sh
@@ -33,7 +33,7 @@ _rt_require_dracut_args "$RAPIDO_DIR/autorun/mpath_local.sh" "$@"
 		   timeout id chown chmod env killall getopt basename" \
 	$DRACUT_RAPIDO_INCLUDES \
 	--add-drivers "virtio_scsi virtio_pci sd_mod" \
-	--modules "bash base systemd systemd-initrd dracut-systemd multipath" \
+	--modules "$DRACUT_RAPIDO_MODULES systemd systemd-initrd dracut-systemd multipath" \
 	$DRACUT_EXTRA_ARGS \
 	$DRACUT_OUT \
 	|| _fail "dracut failed"
--- a/cut/mpath_local.sh
+++ b/cut/mpath_local.sh
@@ -15,7 +15,7 @@
 RAPIDO_DIR="$(realpath -e ${0%/*})/.."
 . "${RAPIDO_DIR}/runtime.vars"
 
-_rt_require_dracut_args "$RAPIDO_DIR/autorun/mpath_local.sh" "$@"
+_rt_require_dracut_args_networkless "$RAPIDO_DIR/autorun/mpath_local.sh" "$@"
 
 # the VM should be deployed with two virtio SCSI devices which share the same
 # backing <file> and <serial> parameters. E.g.
--- a/net-conf/20-vm1-eth0.network
+++ b/net-conf/20-vm1-eth0.network
@@ -1,0 +1,6 @@
+[Match]
+KernelCommandLine=rapido.vm_num=1
+Name=eth0
+
+[Network]
+Address=192.168.155.101/24
--- a/runtime.vars
+++ b/runtime.vars
@@ -216,7 +216,8 @@ _rt_require_dracut_args() {
 	DRACUT_RAPIDO_MODULES="bash base"
 
 	# Default dracut installs
-	DRACUT_RAPIDO_INSTALLS="nohup /usr/lib/systemd/systemd-networkd"
+	DRACUT_RAPIDO_INSTALLS_NET="nohup /usr/lib/systemd/systemd-networkd"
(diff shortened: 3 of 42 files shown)


net-conf/20-vm1-eth0.network
```diff
@@ -0,0 +1,6 @@
+[Match]
+KernelCommandLine=rapido.vm_num=1
```
ddiss, line 2: I wasn't aware of this functionality - nice!

runtime.vars
```diff
@@ -205,10 +205,21 @@ _rt_require_dracut_args() {
...
+				--include $net_conf /etc/systemd/network"
+
+	# Default dracut modules
+	DRACUT_RAPIDO_MODULES="$DRACUT_RAPIDO_MODULES \
+				--modules bash \
+				--modules base \
+				--modules systemd \
+				--modules systemd-initrd \
+				--modules systemd-networkd \
+				--modules dracut-systemd"
```
ddiss, line 222: Adding systemd to all images is going to be quite a hit WRT image size (and likely boot time). I'd hoped that we could run the systemd-networkd binary independently if possible. Also, ideally we only need it for images that require network (i.e. without `_rt_xattr_vm_networkless_set`).

cut/mpath_local.sh
```diff
@@ -33,7 +33,8 @@ _rt_require_dracut_args "$RAPIDO_DIR/autorun/mpath_local.sh" "$@"
 		   timeout id chown chmod env killall getopt basename" \
 	$DRACUT_RAPIDO_INCLUDES \
 	--add-drivers "virtio_scsi virtio_pci sd_mod" \
-	--modules "bash base systemd systemd-initrd dracut-systemd multipath" \
+	$DRACUT_RAPIDO_MODULES \
+	--modules "multipath" \
```
ddiss, line 37: does dracut support multiple `--modules` parameters? I think I'd prefer to use `--modules "$DRACUT_RAPIDO_MODULES multipath"` instead if you're okay with it.
It's inconsistent with `$DRACUT_RAPIDO_INCLUDES`, but double-param `--include` arguments are a special case.

ddiss: > I think I'd prefer to use --modules "$DRACUT_RAPIDO_MODULES multipath" instead if you're okay with it.

...or just keep it as-is if we can get away without the new modules (and instead just add the systemd-networkd binary).

scabrero: Yes, it supports multiple --modules parameters. Ok, I will change it.

runtime.vars
scabrero: I tried adding the binary and it works. I will push an update without adding these modules.

runtime.vars
```diff
@@ -205,10 +205,18 @@ _rt_require_dracut_args() {
...
 				--include $env_src /.profile \
-				--include $emerg_src $emerg_dst"
+				--include $emerg_src $emerg_dst \
+				--include $net_conf /etc/systemd/network"
+
+	# Default dracut modules
+	DRACUT_RAPIDO_MODULES="bash base"
+
+	# Default dracut installs
+	DRACUT_RAPIDO_INSTALLS="nohup /usr/lib/systemd/systemd-networkd"
```
ddiss, line 219: I'd like to only have this installed for images that don't flag themselves as networkless via `_rt_xattr_vm_networkless_set`. Given that networkless is only flagged after image generation, I think for now we should just set:
`DRACUT_RAPIDO_INSTALLS_NET="nohup /usr/lib/systemd/systemd-networkd ip ping"`

and only add `$DRACUT_RAPIDO_INSTALLS_NET` to the `--install` list for images that don't flag themselves networkless.

Alternatively we could add a `_rt_require_dracut_args_networkless` wrapper alongside  `_rt_require_dracut_args` and conditionally set `$DRACUT_RAPIDO_INSTALLS`.

ddiss (changes requested):
This looks like it's almost ready. A couple of things I'd still like to look at:
- documentation should be updated for this new network setup
- we should probably provide a helper to migrate existing rapido.conf IPs to the new config
- some feedback from other rapido users would be nice
  - @luis-henrix can you live with this (minimal) use of `systemd-networkd`? I think it puts us in a better position for supporting more complex network setups, while avoiding extra cludge in rapido.conf . Dropping kernel `ip=` for assignment means we could also finally move towards supporting regular distro kernels (with `KERNEL_SRC` commented out).

Thanks a lot Samuel!

runtime.vars
ddiss: Thinking about this more, I think the `_rt_require_dracut_args_networkless` wrapper would probably be my preference here.

scabrero: Done

vm_autorun.env
```diff
@@ -151,10 +151,18 @@ function _vm_ar_rbd_map
...
+_vm_ar_hostname_set
+
+for f in /sys/class/net/*; do
+	udevadm test $f > /dev/null 2>&1
+done
+
+echo "systemd-network:x:482:482:systemd Network Management:/:/sbin/nologin" \
+       >> /etc/passwd || _fatal
+mkdir /run/systemd/
+nohup /usr/lib/systemd/systemd-networkd &
```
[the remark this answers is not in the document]
scabrero: Done

vm.sh
```diff
@@ -42,28 +42,14 @@ function _vm_start
 	# XXX rapido.conf VM parameters are pretty inconsistent and confusing
 	# moving to a VM${vm_num}_MAC_ADDR or ini style config would make sense
 	local qemu_netdev=""
-	local kern_ip_addr=""
 	if [ -n "$(_rt_xattr_vm_networkless_get ${DRACUT_OUT})" ]; then
 		# this image doesn't require network access
-		kern_ip_addr="none"
 		qemu_netdev="-net none"	# override default (-net nic -net user)
```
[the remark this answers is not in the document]
scabrero: Done

vm.sh
```diff
@@ -42,28 +42,14 @@ function _vm_start
...
 	else
```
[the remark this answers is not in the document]
scabrero: Done

scabrero:
Removing IP_ADDR1 and IP_ADDR2 from autorun scripts won't be so easy as get the address from the interface. The ctdb_cephfs.sh script for example needs both to generate the CTDB nodes file.

ddiss:
@scabrero I'm revisiting your work here, as I need more than two networked VMs for some ldap+automount test cases that I'm running. With openSUSE Leap 15.3 I'm able to run systemd-networkd standalone and it appears to handle static and dhcp configured interfaces quite nicely, along with MAC settings, etc. :partying_face:

I plan on doing more bloat measurements / benchmarks before taking this further, especially given the change in config syntax. I'd hoped that we could additionally use systemd-network configuration for host-side bridge and tap devices, but I haven't yet found a way how we can do that cleanly.

ddiss:
https://github.com/rapido-linux/rapido/pull/173 is a continuation of this PR, so I'm going to close this.

PR #150, opened 2021-02-09
LTP: Add kernel config

pevik:
Required by some tests.

Signed-off-by: Petr Vorel <pvorel@suse.cz>


ddiss (approved):
I would have thought `zcat /proc/config.gz` might be an option, but this also looks okay to me.

ddiss:
Merged as a54ba8970093eee551efe6750a10b9193e5e4b6c (with minor change to quote variable).

PR #154, opened 2021-03-09
cut: drop unnecessary "bash" Dracut module

ddiss:
This change is a straight s/bash base/base/ substitution in cut/*.
The Dracut "base" module already installs the bash binary, so explicitly
adding it via the extra "bash" module is unnecessary.


morbidrsa:
Looks good

PR #162, opened 2021-10-08
trivial: vm error if MAC_ADDR is unset. mpio usage update

ddiss:


luis-henrix (approved):
Although this is a bashism (which I try to avoid) I guess we have already a lot of other bash dependencies on these scripts anyway.  So... meh

luis-henrix (approved):
LGTM

ddiss:
Merged - thanks Luis!

PR #164, opened 2021-12-14
Fastest fixes

morbidrsa:
Two small xfstests fixes, one for all local filesystems and one specific to btrfs on zoned devices.


ddiss (commented):
Merged - thanks Johannes!

PR #167, opened 2021-12-22
autorun/tcmu: properly fork tcmu-runner daemon

ddiss:
Running tcmu-runner as a simple bash background job means that ctrl-c
entries at the shell prompt result in SIGINT delivery and subsequent
tcmu-runner shutdown.
setsid is provided via Dracut's base module, so should always be used
for proper process daemonization.


luis-henrix (approved):
These changes look good to me.

ddiss:
Merged. Thanks Luis!

PR #169, opened 2022-01-07
Initramfs verify mtime

ddiss:


cut/initramfs_verify.sh
```diff
@@ -9,19 +9,32 @@ _rt_require_dracut_args "$RAPIDO_DIR/autorun/initramfs_verify.sh" "$@"
...
+# append as a separate cpio archive.
+if ! grep -q "^# CONFIG_INITRAMFS_PRESERVE_MTIME" "$KERNEL_SRC/.config"; then
+	mkdir -p "${tmp_vdata}/fiod.mtime_chk/2"
+	touch --date=@1641548271 "${tmp_vdata}/fiod.mtime_chk"
+	touch --date=@1641548272 "${tmp_vdata}/fiod.mtime_chk/2"
+	echo -e "fiod.mtime_chk\nfiod.mtime_chk/2" \
+	  | cpio -o -H newc -D "$tmp_vdata"  >> "$DRACUT_OUT" \
+		|| _fail "failed to append mtime_chk archive"
+fi
+
```
luis-henrix, line 39: This doesn't look pretty but I guess there's not an easy way to work around it.  So... meh :-)

luis-henrix (approved):
This looks good.  I've gave it a try with different kernels, and it seems to work fine.

ddiss:
Merged, thanks for the review Luis!

cut/initramfs_verify.sh
ddiss: yeah, not ideal :-) . I figured it's too insignificant to bother fixing in Dracut for now. I might revisit it when I get to https://github.com/dracutdevs/dracut/issues/1662

PR #170, opened 2022-01-09
Allow fstests to run against external devices

ddiss:
This change allows all `fstests_*` runners to use non-zram storage devices. Local block devices (E.g. passed in via `QEMU_EXTRA_ARGS`) must carry a device serial which matches a corresponding xfstests config parameter for which they should be used. 
The Btrfs specific `SCRATCH_DEV_POOL` list parameter can be filled with multiple devices by providing unique devices serials matching a `SCRATCH_DEV*` wildcard.

`fstests_btrfs_zoned` has been left as is for now, as I'm not convinced it makes sense for that specific case.


ddiss:
I've tacked on some some blktests clean ups, as well as some changes which split the Ceph RBD map helper from `vm_autorun.env` into a separate ceph-specific library.

autorun/lib/fstests.sh
```diff
@@ -0,0 +1,77 @@
...
+
+# any dev with a serial number matching an fstests param will be used
+_fstests_devs_provision() {
+	local cfg_path="$1"
+	local i ser
+	declare -A _CFG=(["SCRATCH_DEV"]="" ["SCRATCH_LOGDEV"]="" \
+			 ["SCRATCH_RTDEV"]="" ["TEST_DEV"]="")
+
+	pushd /sys/block
+	for i in */serial; do
```
ddiss, line 24: This is ugly if no block devices with a corresponding `serial` file exist. bash's global `shopt nullglob` behaviour is a mess so I'll need to rework this.

ddiss:
I've force pushed an update which carries the following minor changes:
- rework block device iteration in `_fstests_devs_provision` and `_fstests_devs_pool_provision` to handle absent `serial` entries
- rebase against master branch

autorun/lib/fstests.sh
ddiss: I've turned this into:
```
for i in $(ls /sys/block); do
        ser="$(cat /sys/block/${i}/serial 2>/dev/null)" || continue
        ...
```

autorun/fstests_ext4.sh
```diff
@@ -1,23 +1,12 @@
...
+# SPDX-License-Identifier: (LGPL-2.1 OR LGPL-3.0)
+# Copyright (C) SUSE LLC 2021-2022, all rights reserved.
 
 _vm_ar_env_check || exit 1
 
 set -x
 
-num_devs="2"
-modprobe zram num_devices="${num_devs}" || _fatal "failed to load zram module"
+modprobe zram num_devices="0" || _fatal "failed to load zram module"
```
luis-henrix, line 10: Suggestion: move this zram modprobe (and similar ones in other autorun scripts) into _fstests_devs_zram_setup().  Maybe this function could check /proc/modules to see if it's not loaded yet.  Or, maybe even better, load it if /sys/class/zram-control isn't present, because zram may be built-in.

luis-henrix (changes requested):
I didn't thoroughly test all these commits, but I couldn't find any issues.  Even the only comment I have (related to the zram module load) isn't really major, because the issue was already present in the autorun scripts anyway.
Anyway, awesome stuff!  Thanks, David.

ddiss:
> I didn't thoroughly test all these commits, but I couldn't find any issues. Even the only comment I have (related to the zram module load) isn't really major, because the issue was already present in the autorun scripts anyway. Anyway, awesome stuff! Thanks, David.

Thanks a lot for the review, Luis! I'll force push the minor change to trigger the `"zram hot add failed"` error path here.

ddiss:
New version has the following change squashed in:
``` _fstests_devs_zram_setup() {
        local cfg_var="$1"
-       local zram_num=$(cat /sys/class/zram-control/hot_add) \
-               || _fatal "zram hot add failed"
+       local zram_num

+       zram_num="$(cat /sys/class/zram-control/hot_add)" \
+               || _fatal "zram hot add failed"
        echo "${FSTESTS_ZRAM_SIZE:-1G}" > \
```
I'm still surprised that this is necessary to properly trigger the `"zram hot add failed"` error path, but this is bash we're talking about <sigh>.

luis-henrix:
> I'm still surprised that this is necessary to properly trigger the `"zram hot add failed"` error path, but this is bash we're talking about .

Yeah, it's a bit odd.  The reason is, I think, that `local` will always return 0, and the `cat` execution result doesn't really matter.  Here's from bash(1):

> "The return status is 0 unless local is used outside a function, an invalid name is supplied, or name is a readonly variable."

Anyway, I'm happy with this series so feel free to push it.  As I said, I didn't test it thoroughly (for ex, I didn't tested the rbd changes), but everything looks sane to me.

PR #174, opened 2022-01-30
rdma: use the rdma tool

x2c3z4:
The 5.17-rcX kernel[1] has removed the module parameters, so
have to use the 'rdma' tool.

[1]: https://www.spinics.net/lists/linux-rdma/msg108004.html

Signed-off-by: Li Feng <fengli@smartx.com>

### DIFF
--- a/autorun/nvme_rdma.sh
+++ b/autorun/nvme_rdma.sh
@@ -53,7 +53,7 @@ modprobe zram num_devices="0"
 
 _vm_ar_dyn_debug_enable
 
-echo eth0 > /sys/module/rdma_rxe/parameters/add
+rdma link add rxe0 type rxe netdev eth0 || _fatal
 
 nvmet_cfs="/sys/kernel/config/nvmet/"
 nvmet_subsystem="nvmf-test"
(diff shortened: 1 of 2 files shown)


autorun/nvme_rdma.sh
```diff
@@ -53,7 +53,11 @@ modprobe zram num_devices="0"
 
 _vm_ar_dyn_debug_enable
 
-echo eth0 > /sys/module/rdma_rxe/parameters/add
+if [ -f /sys/module/rdma_rxe/parameters/add ];then
+    echo eth0 > /sys/module/rdma_rxe/parameters/add
+else
+    rdma link add rxe0 type rxe netdev eth0
+fi
```
ddiss, line 60: Thanks a lot for the pull request!
Given that `rdma` should now always be present, can't we use it unconditionally here? Also, please call `_fatal` on error here.

ddiss (changes requested):

autorun/nvme_rdma.sh
x2c3z4: Done.

ddiss (approved):
merging now, thanks!

PR #176, opened 2022-04-14
boot local kernel if KERNEL_SRC is undefined

ddiss:
We're no longer reliant on `CONFIG_IP_PNP=y` for networking, so regular distro kernels can be supported. We currently make quite a few assumptions regarding built-in kernel functionality, but we can iron those things out over time.

This patchset also includes some readme improvements.


pevik:
`Reviewed-by: Petr Vorel <pvorel@suse.cz>` for all commits. I have one note for 3rd commit.

runtime.vars
```diff
@@ -255,53 +258,66 @@ _rt_require_networking() {
...
-	local f="$KERNEL_SRC/.config"
-	local arch_dir boot_dir boot_img qemu_bin
-	QEMU_ARCH_VARS=
+	local arch_dir boot_dir boot_img qemu_bin f grep_cmd
+	if [[ -n $KERNEL_SRC ]]; then
+		_rt_require_conf_dir KERNEL_SRC
+		f="$KERNEL_SRC/.config"
+		grep_cmd="grep -q"
+	else
+		f="/proc/config.gz"	# XXX could use uname info instead?
```
pevik, line 272: If `/proc/config.gz` does not exist, could you please try `/boot/config-$(uname -r)`?

pevik: Actually `uname -m` has all we need, IMHO better to use it.

ddiss: @pevik thanks a lot for the review feedback! I'll go with `/boot/config-$(uname -r)` for now, but will consider making the `uname -m` change in future following further testing.

ddiss:
Merged... I squashed in one further change to use `uname -r` when determining the /boot kernel path instead of the base symlink (e.g. `/boot/vmlinuz`). Otherwise we attempt to boot with differing kernel and kmod versions following host kernel update.

PR #177, opened 2022-04-21
LTP: enable loopback, add dependencies

pevik:
Various LTP improvements.

### DIFF
--- a/cut/ltp.sh
+++ b/cut/ltp.sh
@@ -19,6 +19,14 @@ _rt_require_dracut_args "$RAPIDO_DIR/autorun/ltp.sh" "$@"
 _rt_require_conf_dir LTP_DIR
 _rt_mem_resources_set "2048M"	# 2 vCPUs, 2G RAM
 
+if [[ -n $KERNEL_SRC ]]; then
+	config="${KERNEL_SRC}/.config"
+else
+	config="/boot/config-$(uname -r)"
+fi
+
+[ -f $config ] || _warn "missing kernel config"
+
 "$DRACUT" \
 	--install " \
 		attr awk basename bc blockdev cat chattr chgrp chmod chown cmp cut \
@@ -31,7 +39,7 @@ _rt_mem_resources_set "2048M"	# 2 vCPUs, 2G RAM
 		rmdir sed seq setfattr sort stat strace sync sysctl tac tail tar tc \
 		tee touch tr true truncate uniq unlink vgremove wc which xargs xxd yes" \
 	--include "$LTP_DIR" "$LTP_DIR"  \
-	--include "${KERNEL_SRC}/.config" /.config \
+	--include "$config" /.config \
 	--add-drivers "loop" \
 	--modules "base" \
 	"${DRACUT_RAPIDO_ARGS[@]}" \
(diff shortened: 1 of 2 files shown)


cut/ltp.sh
```diff
@@ -19,18 +19,27 @@ _rt_require_dracut_args "$RAPIDO_DIR/autorun/ltp.sh" "$@"
...
-	--install "tail blockdev ps rmdir resize dd grep find df mkfs which \
-		perl awk bc touch cut chmod true false unlink mktemp getfattr \
-		setfattr attr killall hexdump sync id sort uniq date expr tac \
-		diff head dirname seq basename tee egrep yes dmsetup chattr \
-		lsattr cmp stat hostname getconf md5sum od wc tr xargs sysctl \
-		link truncate quota quotacheck quotaon vgremove chgrp du fgrep \
-		pgrep pkill tar rev kill fdformat ldd free losetup chown sed \
-		cat lsmod ip ping tc \
-		${LTP_DIR}/bin/* ${LTP_DIR}/testcases/bin/*" \
+	--install "${LTP_DIR}/bin/* ${LTP_DIR}/testcases/bin/* ${LTP_DIR}/testcases/data/ \
```
ddiss, line 31: Is `${LTP_DIR}/testcases/data/` needed here? We already have a separate `--include "$LTP_DIR" "$LTP_DIR"` parameter below, which should provide the `testcases/data/` contents. The extra `--install` entries are normally only needed to ensure that shared object libraries are included for any binaries - does `testcases/data/` contain binaries?

ddiss (approved):
Looks good - thanks!
I'll wait to hear whether the explicit `testcases/data` install is needed before merging.

cut/ltp.sh
pevik: The only shared libraries are some testing libraries:
```
/opt/ltp/testcases/data/ldd01/lddfile2.obj.so
/opt/ltp/testcases/data/ldd01/lddfile3.obj.so
/opt/ltp/testcases/data/ldd01/lddfile1.obj.so
/opt/ltp/testcases/data/ldd01/lddfile4.obj.so
/opt/ltp/testcases/data/ldd01/lddfile5.obj.so
```
But even these are installed.
BTW `ls -R /opt/ltp/ |wc -l` gives a same number thus kept only `--include "$LTP_DIR" "$LTP_DIR"`.

ddiss: any shared libraries under `LTP_DIR` will indeed be installed via the `--include "$LTP_DIR" "$LTP_DIR"`, but the `bin` paths should still be listed under `--install` to ensure that *system* shared libraries get pulled in. E.g. ltp can be compiled with libnuma support, in which case we get:
```
david@balakirew:~/ltp> ldd inst_dest/opt/ltp/testcases/bin/get_mempolicy01
        linux-vdso.so.1 (0x00007ffc00ff4000)
        libnuma.so.1 => /usr/lib64/libnuma.so.1 (0x00007fc61d139000)
        libc.so.6 => /lib64/libc.so.6 (0x00007fc61cd44000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fc61d345000)
```

Dracut will only pull `libnuma.so.1` into the initramfs image if `testcases/bin/get_mempolicy01` is covered by an `--install` path.
I'll merge your change with the `--install "${LTP_DIR}/bin/* ${LTP_DIR}/testcases/bin/* \` line left as-is.

ddiss:
pushed with some minor changes applied.

PR #178, opened 2022-04-22
Introduce $KCONFIG: path for kconfig file (DRY)

pevik:
Currently cut/ltp.sh and cut/initramfs_verify.sh require kconfig path (possibly more in the future), therefore put it into runtime.vars.

Also warn if file missing.

NOTE: `$KCONFIG` is common variable for config on kernel testing tools (LTP and many others).


ddiss (commented):
Thanks for the submission, but I think I'd prefer to avoid the extra global variable. We could add something like `_rt_require_kconfig()`, but given that there are only two consumers, my preference would be to keep things as-is.

At this stage we only have

pevik:
Sure, np.

PR #179, opened 2022-04-26
uring_btrfs: increase memory and bring up loopback

ddiss:
I'm seeing OOMs with liburing 2.1 on SLE15-SP4, so increase mem to 2G.
loopback is needed for the send_recvmsg test.

### DIFF
--- a/autorun/uring_btrfs.sh
+++ b/autorun/uring_btrfs.sh
@@ -1,16 +1,6 @@
 #!/bin/bash
-#
-# Copyright (C) SUSE LLC 2021, all rights reserved.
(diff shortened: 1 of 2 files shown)


pevik:
`Signed-off-by: Petr Vorel <pvorel@suse.cz>`

nit: isn't loopback setup separated thing (could be in separate commit).

ddiss:
Merged - thanks for the review, Petr!

> nit: isn't loopback setup separated thing (could be in separate commit).

Yeah, I should probably have split them. Will keep this in mind for next time.

PR #180, opened 2022-05-04
runtime: use dracut cmdline hook instead of emergency

ddiss:
00-rapido-init.sh is currently installed as a Dracut emergency hook,
which gets invoked via:
```
  [ -z "$root" ] && die "No or empty root= argument"
  # die() later invokes emergency_shell() which invokes rapido-init
```
This means that we see a number of worrisome error messages before
rapido starts, e.g.
```
  [    0.789194] Run /init as init process
  [    0.860096] dracut: openSUSE Leap-15.3
  [    0.886398] dracut: FATAL: No or empty root= argument
  [    0.887587] dracut: Refusing to continue

  [    0.891465] dracut Warning:
  dracut Warning:

  Rapido: starting /rapido_autorun/00-simple_example.sh
```
With this change, 00-rapido-init.sh is invoked as a dracut cmdline hook
which is executed before "root=" cmdline processing, resulting in less
noise on the console, e.g.:
```
  [    0.706658] Run /init as init process
  [    0.774859] dracut: openSUSE Leap-15.3
  Rapido: starting /rapido_autorun/00-simple_example.sh
```
While testing with systemd I found that we also should be installing
vm_autorun.env under /etc/profile instead of /.profile; the latter
is reliant on a '/' home directory path.

### DIFF
--- a/runtime.vars
+++ b/runtime.vars
@@ -193,13 +193,14 @@ _rt_require_dracut_args() {
 	# destination filename.
 	local init_src=$1 init_dst kver
 
-	# specify core init scripts responsible for starting autorun
+	# Specify core init scripts responsible for starting autorun.
+	# The Dracut "cmdline" hook is executed prior to root parameter parsing.
 	local env_src="$RAPIDO_DIR/vm_autorun.env"
-	local emerg_src="$RAPIDO_DIR/autorun/00-rapido-init.sh"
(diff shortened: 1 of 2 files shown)


ddiss:
This one works nicely for me on both Tumbleweed and Leap. I plan on merging it today.

PR #183, opened 2022-06-01
cut/fstests_btrfs: add ext dependencies for convert

ddiss:
The fstests "convert" group requires a bunch of ext user space
dependencies as well as the ext4 kernel module.

I've left fstests_btrfs_zoned as-is, as btrfs-convert probably doesn't
make sense there until ext4 is zone aware.


ddiss:
@morbidrsa can you confirm that `fstests_btrfs_zoned` can be left as-is?

morbidrsa:
> @morbidrsa can you confirm that `fstests_btrfs_zoned` can be left as-is?

Yes, ext4 will probably never work with zoned devices, so we don’t have to care here.

PR #185, opened 2022-06-27
runtime: add and load virtio-rng / 9p modules on demand

ddiss:
If the virtio-rng and 9p virtfs modules aren't built-in (e.g. when
booting a distro kernel with KERNEL_SRC unset), then the modules need
to be explicitly added to the initramfs image and loaded on boot.

Do so if the corresponding virtio-rng or 9p functionality is requested
via QEMU_EXTRA_ARGS and / or VIRTFS_SHARE_PATH respectively.

### DIFF
--- a/runtime.vars
+++ b/runtime.vars
@@ -191,7 +191,7 @@ _rt_require_dracut_args() {
 	# Multiple scripts can be specified and will be run in the order of
 	# parameters. This order is controlled by prepending an index to the
 	# destination filename.
-	local init_src=$1 init_dst kver
+	local init_src=$1 init_dst kver kmods=()
 
 	# Specify core init scripts responsible for starting autorun.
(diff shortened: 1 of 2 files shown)


pevik:
@ddiss Good idea, I'd just put functions in some single place and reuse it (DRY)

ddiss:
> 

Thanks for the feedback. At the moment we don't really offer a clean way to share this sort of code between host and VM, so my preference would be to merge as-is if you don't feel too strongly about it.

pevik:
Sorry for a delay. Sure, go ahead. It's just a second PR when you copy paste. Sooner or later it'd be nice to have way to avoid that.

`Reviewed-by: Petr Vorel <pvorel@suse.cz>`

ddiss:
Merged as 745a76e48aae2d6ae79feed7f8be9fa78b96092f .

PR #186, opened 2022-08-15
conf: Add default value for LTP_DIR

pevik:
Most of users install LTP to the default prefix /opt/ltp,
therefore enable it by default (saves the need to change it for most of
users).


ddiss:
Merged as cb29d8245b1f925bf75078df73075a68cd421475 - thanks!

PR #190, opened 2022-10-24
LTP small improvements

pevik:
* cut/ltp.sh: Add more tools
* ltp: Use SPDX
* autorun/ltp.sh: cd to LTP tests directory
* autorun/ltp.sh: export LTPROOT

### DIFF
--- a/cut/ltp.sh
+++ b/cut/ltp.sh
@@ -1,16 +1,6 @@
 #!/bin/bash
-#
-# Copyright (C) SUSE LINUX GmbH 2019, all rights reserved.
-#
-# This library is free software; you can redistribute it and/or modify it
-# under the terms of the GNU Lesser General Public License as published
-# by the Free Software Foundation; either version 2.1 of the License, or
-# (at your option) version 3.
-#
-# This library is distributed in the hope that it will be useful, but
-# WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
-# or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public
-# License for more details.
+# SPDX-License-Identifier: (LGPL-2.1 OR LGPL-3.0)
+# Copyright (C) SUSE LLC 2019-2022, all rights reserved.
 
 RAPIDO_DIR="$(realpath -e ${0%/*})/.."
 . "${RAPIDO_DIR}/runtime.vars"
--- a/cut/ltp.sh
+++ b/cut/ltp.sh
@@ -25,17 +25,18 @@ else
 	config="/boot/config-$(uname -r)"
 fi
 
+fses=(btrfs exfat ext2 ext3 ext4 fuse ntfs vfat xfs)
+
 "$DRACUT" \
 	--install " \
-		attr awk basename bc blockdev cat chattr chgrp chmod chown cmp cut \
-		date dd df diff dirname dmsetup du egrep expr false fdformat fdisk \
-		fgrep find free gdb getconf getfattr grep head hexdump hostname id ip \
-		kill killall ldd link losetup lsattr lsmod ltrace md5sum mkfs mkfs.bfs \
-		mkfs.btrfs mkfs.cramfs mkfs.ext2 mkfs.ext3 mkfs.ext4 mkfs.fat mkfs.jfs \
-		mkfs.minix mkfs.msdos mkfs.ntfs mkfs.vfat mkfs.xfs mktemp od parted \
-		perl pgrep ping ping6 pkill ps quota quotacheck quotaon resize rev \
-		rmdir sed seq setfattr sort stat strace sync sysctl tac tail tar tc \
-		tee touch tr true truncate uniq unlink vgremove wc which xargs xxd yes \
+		attr awk basename bc blockdev cat chattr chgrp chmod chown cmp cut date
+		dd df diff dirname dmsetup du egrep exportfs expr false fdformat fdisk
+		fgrep find free gdb getconf getfacl getfattr grep head hexdump hostname
+		id ip kill killall ldd link losetup lsattr lsmod ltrace md5sum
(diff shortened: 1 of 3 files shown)


cut/ltp.sh
```diff
@@ -18,11 +18,10 @@ fi
...
 	--install " \
 		attr awk basename bc blockdev cat chattr chgrp chmod chown cmp cut \
-		date dd df diff dirname dmsetup du egrep expr false fdformat fdisk \
+		date dd df diff dirname dmsetup du egrep exportfs expr false fdformat fdisk \
 		fgrep find free gdb getconf getfattr grep head hexdump hostname id ip \
-		kill killall ldd link losetup lsattr lsmod ltrace md5sum mkfs mkfs.bfs \
-		mkfs.btrfs mkfs.cramfs mkfs.ext2 mkfs.ext3 mkfs.ext4 mkfs.fat mkfs.jfs \
-		mkfs.minix mkfs.msdos mkfs.ntfs mkfs.vfat mkfs.xfs mktemp od parted \
+		kill killall ldd link losetup lsattr lsmod ltrace md5sum $(ls /sbin/mkfs*) \
+		mktemp $(ls /sbin/mount*) od parted \
```
ddiss, line 24: Thanks a lot for the submission, Petr!
The changes look fine, but I'd prefer to avoid these wildcards, just in case someone has a private `/sbin/mount-my-hdd-with-luks-password` script or something.
My preference would be something like:
```
fses=(btrfs ext2 ext3 ext4 ...)
dracut --install "... ${fses[*]/#/mount.} ${fses[*]/#/mkfs.} ..."
```

ddiss (changes requested):

cut/ltp.sh
pevik: If you're ok, that I'll include `fuse`, there is only `mount.fuse`, but not `mkfs.fuse`, thus extra error will be produced.

pevik:
@ddiss  I'm sorry for many changes, fixed few problems. Could you please have a second look?

cut/ltp.sh
ddiss: good point. in that case I think it'd make sense to go back to listing them all separately, without globbing (sorry for the run-around).

cut/ltp.sh
```diff
@@ -25,18 +15,18 @@ else
...
-		attr awk basename bc blockdev cat chattr chgrp chmod chown cmp cut \
-		date dd df diff dirname dmsetup du egrep expr false fdformat fdisk \
-		fgrep find free gdb getconf getfattr grep head hexdump hostname id ip \
-		kill killall ldd link losetup lsattr lsmod ltrace md5sum mkfs mkfs.bfs \
-		mkfs.btrfs mkfs.cramfs mkfs.ext2 mkfs.ext3 mkfs.ext4 mkfs.fat mkfs.jfs \
-		mkfs.minix mkfs.msdos mkfs.ntfs mkfs.vfat mkfs.xfs mktemp od parted \
-		perl pgrep ping ping6 pkill ps quota quotacheck quotaon resize rev \
-		rmdir sed seq setfattr sort stat strace sync sysctl tac tail tar tc \
-		tee touch tr true truncate uniq unlink vgremove wc which xargs xxd yes \
-		${LTP_DIR}/bin/* ${LTP_DIR}/testcases/bin/*" \
```
ddiss, line 39: why did you drop the `LTP_DIR/...` installs? They're likely still needed for pulling in the shared object dependencies. IIRC, we discussed this with the initial merge.

pevik: I guess you're talking about https://github.com/rapido-linux/rapido/pull/85#discussion_r280192054. You were discussing libraries. There are no shared libraries, there are only statically linked libraries:

```
$ find libs/ |grep -e '\.so$' -e '\.a'
libs/libltpipc/libltpipc.a
libs/libltpnewipc/libltpnewipc.a
libs/libltpnuma/libltpnuma.a
libs/libltpsigwait/libltpsigwait.a
libs/libltpswap/libltpswap.a
libs/libltpuinput/libltpuinput.a
libs/libltpvdso/libltpvdso.a
```

Everything installed in `/opt/ltp/testcases/bin` is included with the current setup. Testing `rt_sigtimedwait01`, which uses `libs/libltpsigwait/libltpsigwait.a` and it works with current setup. `msgctl01` which uses `libs/libltpnewipc/libltpnewipc.a` and it also works. And the reason is really that everything is statically compiled.

There are very few dynamic library dependencies used for 0.5% of LTP tests - e.g. libmnl-devel
But with the current setup it works in initramfs:
```
# ldd ./route-change-netlink | grep -i libmnl
        libmnl.so.0 => /usr/lib64/libmnl.so.0 (0x00007f9b26b2f000)
```
Therefore I'd suggest to keep it. But if you insist to keep it, I'll put it back.

ddiss: > I guess you're talking about [#85 (comment)](https://github.com/rapido-linux/rapido/pull/85#discussion_r280192054). You were discussing libraries.

That's the one :-)
...
> There are very few dynamic library dependencies used for 0.5% of LTP tests - e.g. libmnl-devel But with the current setup it works in initramfs:
> 
> ```
> # ldd ./route-change-netlink | grep -i libmnl
>         libmnl.so.0 => /usr/lib64/libmnl.so.0 (0x00007f9b26b2f000)
> ```

These sorts of LTP-external library dependencies are what I'm concerned about. FWIW, `cut/ltp.sh` carries `--install ip`, which is why Dracut pulls in `libmnl.so.0` automatically.

> Therefore I'd suggest to remove `${LTP_DIR}/bin/* ${LTP_DIR}/testcases/bin/*` (as it's in the current code). But if you insist to keep it, I'll put it back. Or give me an example what else could be "shared object dependency".

Even if LTP has very few external dependencies, I'd still prefer to keep using the `dracut --install` shared-object dependency checks to pull in what's needed.

pevik: OK, replaced with commit, which added just directories into `$PATH`.

ddiss (approved):

Issue #192, opened 2022-11-22
fstests: external virtio block devices aren't detected with CONFIG_VIRTIO_BLK=m


ddiss:
The `virtio_blk` kernel module needs to be loaded prior to calling `_fstests_devs*_provision()`, otherwise external virtio block devices won't be detected.

PR #193, opened 2022-11-24
fstests: add and load virtio_blk kernel module

ddiss:
This is required when using host-provided test/scratch block devices (feature added via https://github.com/rapido-linux/rapido/pull/170) on kernels built with CONFIG_VIRTIO_BLK=m. Rapido kconfigs in kernel/ specify CONFIG_VIRTIO_BLK=y, so this is mostly just for stock distro kernels.

cut/fstests_xfs.sh is updated to also pull in loop, scsi_debug and dm-log-writes kernel modules, which some tests desire.

Fixes: https://github.com/rapido-linux/rapido/issues/192

### DIFF
--- a/autorun/fstests_btrfs.sh
+++ b/autorun/fstests_btrfs.sh
@@ -7,6 +7,7 @@ _vm_ar_env_check || exit 1
 set -x
 
 [ -n "$BTRFS_PROGS_SRC" ] && export PATH="${PATH}:${BTRFS_PROGS_SRC}"
+modprobe virtio_blk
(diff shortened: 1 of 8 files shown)


ddiss:
An alternative to unconditionally adding the `virtio-blk` driver would be to check `QEMU_EXTRA_ARGS` for `-device virtio-blk` before doing so, but the kernel module itself is quite small so I decided to keep it simple.

pevik (approved):
`Reviewed-by: Petr Vorel <pvorel@suse.cz>`

ddiss:
pushed - thanks Petr!

PR #200, opened 2023-01-27
tools/qemu_gdb: add kaslr hint and load modules on startup

ddiss:
Kernel address space layout randomization (CONFIG_RANDOMIZE_BASE=y) causes problems for gdb, so add a hint about how to disable it. Run lx-symbols on startup so that symbols are loaded for kernel modules.


pevik (approved):
```
Reviewed-by: Petr Vorel <petr.vorel@gmail.com>
```
+1, good hint.

PR #201, opened 2023-02-14
cut/cifs: add cmac kernel module

ddiss:
Needed for SMB3 connections. Also add the gcm kernel module dependency which was missing from cut/fstests_cifs.sh .


pevik (approved):
Obviously correct.
```
```

PR #206, opened 2023-03-17
fstests_ntfs3 and ksmbd: add cut and autorun scripts

ddiss:
This makes use of the mkfs.ntfs binary from the ntfs-3g project.


pevik (approved):
LGTM

PR #214, opened 2023-06-27
cut/kselftest: add zram test dependencies

ddiss:
Require ext2 and zram kernel modules as well as a few extra userspace binaries which are invoked by the tests scripts.

### DIFF
--- a/cut/kselftest.sh
+++ b/cut/kselftest.sh
@@ -34,8 +34,10 @@ _rt_require_conf_dir KSELFTEST_DIR
 # details
 "$DRACUT" \
 	--install "awk basename cut date dirname echo expr fmt grep head id
-	           lscpu ps realpath rmdir sort uniq wc" \
+	           lscpu ps realpath rmdir sort uniq wc resize seq dd bc
+	           mkswap swapon swapoff mkfs.ext4 which" \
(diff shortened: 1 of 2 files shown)


ddiss:
@Werkov I expect there are going to be many more dependencies to be added as we support more and more components, but we might as well start somewhere.

ddiss:
v2:
- added which binary
- switched to ext4 instead of ext2. The latter is only used as fallback if `which mkfs.ext4` fails
- added extra minor change to disable dracut's `hardlink` call

I plan on pushing in the coming days

ddiss:
I've merged this. Quite a bit more work is needed to expand rapido support for kselftest targets. This is just a start.

Werkov:
FTR, there is also a floating idea of packaging kselftest (RPM, openSUSE(?)) which could offload some of the deps tracking from rapido (although there's non-zero impedance between RPMs and dracut modules).

ddiss:
> FTR, there is also a floating idea of packaging kselftest (RPM, openSUSE(?)) which could offload some of the deps tracking from rapido (although there's non-zero impedance between RPMs and dracut modules).

I saw that raised, thanks. I'd be happy to also support locally installed kselftest if it does end up in a package - that'd then just be a matter of checking for / using the package paths if `KSELFTEST_DIR` is unset.

PR #222, opened 2023-08-22
docs: add animated SVG as demo screencast

ddiss:
The demo cuts and boots simple-example with host kernel and a custom 6.5.0-rc7 based kernel.
The termtosvg output was (and still is) quite bloated. I've done a few things to try to minify it:
- remove unused colors
- move css class colors to rect and text types to avoid explicit class naming
- rename "background" class "bg" to shorten explicit names
- merge separate style sections

I've also sped-up playback (83326ms -> 63326ms animation-duration) to make it less painful to watch.

A link to the raw SVG is added to readme.md, but it will only work after the changes are merged into the master branch. I chose a link instead of embedding the SVG to avoid distraction and unnecessary CPU/memory use on the project landing page.


ddiss:
The raw demo SVG can be viewed at https://raw.githubusercontent.com/rapido-linux/rapido/f8e032d8aabdd31887366e890914cc3e902884f0/docs/demo.svg

PR #226, opened 2023-09-01
runtime: fix _rt_write_ceph_config parsing

ddiss:
The value=(...) bash array syntax doesn't match $value usage, and breaks parsing of, e.g.:
  [global]
    mon host =  [v2:192.168.159.1:40716,v1:192.168.159.1:40717]


pevik (approved):
```
Fixes: 29f3d18 ("runtime.vars: add support for configuration parsing using ceph-conf")
```

ddiss:
Thanks for the review, Petr. Pushed as 5f2c4a6e8405a7da79b99644a51601fe4454f350 .

PR #227, opened 2023-09-01
autorun/ltp: Export `KCONFIG_PATH`

pevik:
Kernel config in rapido is in nonstandard path, point the correct location via `KCONFIG_PATH` environment variable so that user does not have to do it manually.

Also move `KCONFIG_PATH` to `rapido.conf.example`, so that it does not have to be defined on both `autorun/ltp.sh` and `cut/ltp.sh`.

Fixes: db9be25 ("cut/ltp: Install kernel config when local kernel booted")

### DIFF
--- a/autorun/ltp.sh
+++ b/autorun/ltp.sh
@@ -17,6 +17,7 @@ for ug in nobody bin daemon; do
 done
 
 export CREATE_ENTRIES=0
+export KCONFIG_PATH=/.config
 export LTPROOT="$LTP_DIR"
 export PATH="$LTP_DIR:$LTP_DIR/bin:$LTP_DIR/testcases/bin:$PATH"
 


ddiss (changes requested):
`rapido.conf[.example]` is generally reserved for **user** configurable parameters, but IIUC there's generally no reason for a user to change `KCONFIG_PATH`. My preference would be to just use the hardcoded path in autorun, i.e. `export KCONFIG_PATH="/.config"`.

If you really want to share the kconfig path variable privately between cut and autorun, then you could consider adding the `export KCONFIG_PATH=...` logic to a temp file which gets added as first parameter to `_rt_require_dracut_args` (and sourced first when the VM boots). See cut/cephfs.sh as an example.

pevik:
Thanks! I just exported the variable in autorun file.

ddiss:
merged - thanks!

PR #230, opened 2023-12-09
Add net test

acooks:
Thank you for creating and maintaining rapido!

It's been really useful for testing a network device driver I'm working on and I would like to share the changes I've made.

### DIFF
--- a/cut/net_test.sh
+++ b/cut/net_test.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+# SPDX-License-Identifier: (LGPL-2.1 OR LGPL-3.0)
+
+RAPIDO_DIR="$(realpath -e ${0%/*})/.."
+. "${RAPIDO_DIR}/runtime.vars"
+
+[[ -d "${KERNEL_SRC}" ]] || \
+  _fail "specify a valid KERNEL_SRC in rapido.conf"
+
+[[ -d "$KERNEL_INSTALL_MOD_PATH" ]] || \
+  _fail "specify a valid KERNEL_INSTALL_MOD_PATH in rapido.conf"
+
+[[ -n ${NET_TEST_KMOD} ]] || \
+  _fail "specify a kernel module to test in rapido.conf"
+
+[[ -d "$NET_TEST_SUITE" && -x "$NET_TEST_SUITE/start.sh" ]] || \
+  _fail "specify the test suite directory containing start.sh"
+
+_rt_require_dracut_args "$RAPIDO_DIR/autorun/net_test.sh" "$@"
+_rt_cpu_resources_set "2"
+_rt_mem_resources_set "256M"
+
+NET_TEST_TOOLS="\
+	bridge \
+	dhclient \
+	dirname basename realpath \
+	dmesg \
+	ethtool \
+	grep \
+	hostname \
+	ip \
+	iperf3 \
+	lspci \
+	nc \
+	ping \
+"
+
+# ideally dracut would fail if the specified module cannot be found,
+# but it exits with 0 (tested version 059-15.fc39)
+"$DRACUT" \
+	--install "${NET_TEST_TOOLS}" \
+	--add-drivers "af_packet ${NET_TEST_KMOD}.ko" \
+	--modules "base" \
+	--include "${KERNEL_SRC}/samples/pktgen" "pktgen" \
+	--include "${NET_TEST_SUITE}" "driver-tests" \
+	"${DRACUT_RAPIDO_ARGS[@]}" \
+	"$DRACUT_OUT" || _fail "dracut failed"
--- a/autorun/net_test.sh
+++ b/autorun/net_test.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+# SPDX-License-Identifier: (LGPL-2.1 OR LGPL-3.0)
+
+# The tn40xx-test directory and test scripts will be copied
+# by the cut script, from the tn40xx-driver source,
+# as defined in the rapido.conf file.
+
+_vm_ar_env_check || exit 1
+_vm_ar_dyn_debug_enable
+
+ip link set lo up
+
+/driver-tests/start.sh ${NET_TEST_DEV_VM} ${NET_TEST_KMOD}
--- a/docs/network_device_driver_guide.md
+++ b/docs/network_device_driver_guide.md
@@ -1,0 +1,52 @@
+This is an example config file for passthrough network device driver testing.
+
+
+All these variables would likely need to be set according to the developer environment.
+
+
+    #
+    # Path to Linux kernel source. Prior to running "rapido cut", the kernel
+    # should be built, with modules installed (see KERNEL_INSTALL_MOD_PATH below).
+    # e.g. KERNEL_SRC="/home/me/linux"
+    #KERNEL_SRC=""
+
+    # Path to compiled kernel modules.
+    # If left blank, Dracut will use its default (e.g. /lib/modules) search path.
+    # A value of "${KERNEL_SRC}/mods" makes sense when used alongside
+    # "INSTALL_MOD_PATH=./mods make modules_install" during kernel compilation.
+    # e.g. KERNEL_INSTALL_MOD_PATH="${KERNEL_SRC}/mods"
+    #KERNEL_INSTALL_MOD_PATH=""
+
+    # The name of the kernel module under test.
+    # Passed to the VM as a kernel parameter for use in the test suite.
+    # e.g. NET_TEST_KMOD="tn40xx"
+    #NET_TEST_KMOD=""
+
+    # The location of the test suite to add into the VM.
+    # e.g. NET_TEST_SUITE="/home/me/tn40xx-driver/test"
+    #NET_TEST_SUITE=""
+
+    # The host PCI address of the network peripheral for the driver under test.
+    # This device must use the vfio-pci stub driver in the host, to be passed to
+    # the VM.
+    # e.g. NET_TEST_DEV_HOST="07:00.0"
+    #NET_TEST_DEV_HOST=""
+
+    # The PCI address of the peripheral in the VM.
+    # Passed to the VM as a kernel parameter and used in the test suite to identify
+    # the peripheral for the driver under test.
+    # e.g. NET_TEST_DEV_VM="0000:00:03.0"
+    #NET_TEST_DEV_VM=""
+
+    # Enable IOMMU and pass the PCIe network peripheral into the VM.
+    # e.g.
+    # QEMU_EXTRA_ARGS="\
+    #  -M q35,accel=kvm,kernel-irqchip=split \
+    #  -cpu host \
+    #  -nographic\
+    #  -device intel-iommu,intremap=on,caching-mode=on,device-iotlb=on \
+    #  -device virtio-rng-pci \
+    #  -device vfio-pci,host=$NET_TEST_DEV_HOST \
+    # "
+    #QEMU_EXTRA_ARGS=""
+
--- a/rapido.conf.example
+++ b/rapido.conf.example
@@ -45,6 +45,27 @@ TAP_USER=""
 # e.g. VIRTFS_SHARE_PATH="/tmp/rapido/"
 #VIRTFS_SHARE_PATH=""
 
+# The name of the kernel module under test.
+# Passed to the VM as a kernel parameter for use in the test suite.
+# e.g. NET_TEST_KMOD="tn40xx"
+#NET_TEST_KMOD=""
+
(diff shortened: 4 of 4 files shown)


rapido.conf.example-net_test
```diff
@@ -0,0 +1,360 @@
+# Path to Linux kernel source. Prior to running "rapido cut", the kernel
+# should be built, with modules installed (see KERNEL_INSTALL_MOD_PATH below).
+# If KERNEL_SRC is not set then rapido will attempt to boot the local kernel
+# present under /boot/.
+#
+# e.g. KERNEL_SRC="/home/me/linux"
+#KERNEL_SRC=""
+KERNEL_SRC="/home/acooks/linux/"
```
pevik, line 8: I suppose @ddiss will give more feedback. But this is certainly wrong.

rapido.conf.example-net_test
```diff
@@ -0,0 +1,360 @@
...
+ -M q35,accel=kvm,kernel-irqchip=split \
+ -cpu host \
+ -nographic\
+ -device intel-iommu,intremap=on,caching-mode=on,device-iotlb=on \
+ -device virtio-rng \
+ -device vfio-pci,host=$NET_TEST_DEV_HOST \
+"
+
+
+
```
pevik, line 76: Why so many blank lines?

rapido.conf.example-net_test
```diff
@@ -0,0 +1,360 @@
+# Path to Linux kernel source. Prior to running "rapido cut", the kernel
```
pevik, line 1: Copy paste whole `rapido.conf.example` with minimal changes looks wrong to me. Why not just add new section to it?

acooks: I added a complete working example config to show exactly how I'm using the software.

It is meant to help you, the reviewer understand how many things need to be configured and how, because it's not entirely obvious what variables need to be configured and how, when given only the four new variables by themselves.

It is clearly not ready to merge as is, and I was hoping a reviewer would understand the complexity issue and set a direction fixing it. Perhaps there is some value in providing more example config files that show all variables that are relevant to a particular use case and none of the other varables.

How would you proivde the user instructions and avoid adding complexity into the example config?

ddiss:
Thanks for the PR!
As @pevik indicated, this appears to add quite a bit of extra complexity to rapido.conf. Hopefully some of it can be avoided by reworking the changes a little:
* your use case appears to target tn40xx-driver development, so I think it'd be nice to have tn40xx specific cut/autorun scripts instead of generic net-test
  * `cut/tn40xx` could then pull in the driver and test suite from e.g. `TN40XX_DRIVER_SRC`
  * `autorun/tn40xx` could run `start.sh` directly, or let users run it (with flexible parameters) via `rapido cut -x "./start.sh ..." tn40xx` 
* `rapido.conf` is included by VMs on boot, so the `QEMU_EXTRA_KERNEL_PARAMS` changes shouldn't be needed
  * ideally `NET_TEST_DEV_VM` could also be assumed / autodetected by `autorun/tn40xx`
* I appreciate that you've put together a standalone example config for your use case. I think it would be best suited to live under e.g. `docs/tn40xx_guide.md` similar to the Ceph guide

I'll add a few other minor style comments inline...

cut/net_test.sh
```diff
@@ -0,0 +1,56 @@
...
+	dirname basename realpath \
+	dmesg \
+	ethtool \
+	grep \
+	hostname \
+	ip \
+	iperf3 \
+	lspci \
+	nc \
+	ping \
```
ddiss, line 29: please put these after `--install` directly, rather than using a separate variable.

cut/net_test.sh
```diff
@@ -0,0 +1,56 @@
...
+
+[[ -d "${KERNEL_SRC}" ]] || \
+  _fail "specify a valid KERNEL_SRC in rapido.conf"
+
+[[ -d "$KERNEL_INSTALL_MOD_PATH" ]] || \
+  _fail "specify a valid KERNEL_INSTALL_MOD_PATH in rapido.conf"
+
+
+[[ "x" != "x${NET_TEST_KMOD}" ]] || \
+  _fail "specify a kernel module to test in rapido.conf"
```
ddiss, line 40: `[[ -n $NET_TEST_KMOD ]]` should work fine for this, but as mentioned, hopefully this can be hardcoded as a tn40xx specific kernel module.

cut/net_test.sh
```diff
@@ -0,0 +1,56 @@
...
+
+
+# ideally dracut would fail if the specified module cannot be found,
+# but it exits with 0 (tested version 059-15.fc39)
+"$DRACUT" \
+	--install "${NET_TEST_TOOLS}" \
+	--add-drivers "af_packet ${NET_TEST_KMOD}.ko" \
+	--modules "base" \
+	--include "${KERNEL_SRC}/samples/pktgen" "pktgen" \
+	--include "${NET_TEST_SUITE}" "driver-tests" \
```
ddiss, line 54: If possible, please use the same host source and VM destination paths. The reason for doing so is that `dracut --install` can then also be used to pull in any additional shared object dependencies; `dracut --install <bin>` parses `ldd <bin>` output for dependencies while `dracut --include` ignores it.

ddiss (changes requested):

acooks:
Thank you for the quick review and feedback!

>     * your use case appears to target tn40xx-driver development, so I think it'd be nice to have tn40xx specific cut/autorun scripts instead of generic net-test
>       

Are you sure you'd want to be so specific about the driver?

I started from the very specific case, but thought it was too specific to upstream, because it wouldn't be useful to anyone else. However, the generic case for testing drivers for passed-through devices seems like more than just a once-off need.

For example, it might also be useful for these similar out-of-mainline network drivers:
https://github.com/morrownr/8814au
https://github.com/Mange/rtl8192eu-linux-driver/
https://github.com/lwfinger/rtl8192du

The goal for these drivers is probably to cease stand-alone development and be superseded by a mainline driver, but even then, the rapido developer workflow that enables passing the peripheral and driver into a VM is so much better than any of the alternatives I've seen (and not specific to network devices, or pci devices, or even device drivers).

cut/net_test.sh
acooks: It's great that dracut is smart enough to do that.  I figured I'd put all the standard utilities with discoverable shared object dependencies in the install parameter, and create a conventional place to put the custom test suite, which may have non-discoverable shared objects and unknowable other cruft.

ddiss:
> Thank you for the quick review and feedback!

No worries :-)

> > ```
> > * your use case appears to target tn40xx-driver development, so I think it'd be nice to have tn40xx specific cut/autorun scripts instead of generic net-test
> > ```
> 
> Are you sure you'd want to be so specific about the driver?
> 
> I started from the very specific case, but thought it was too specific to upstream, because it wouldn't be useful to anyone else. However, the generic case for testing drivers for passed-through devices seems like more than just a once-off need.
> 
> For example, it might also be useful for these similar out-of-mainline network drivers: https://github.com/morrownr/8814au https://github.com/Mange/rtl8192eu-linux-driver/ https://github.com/lwfinger/rtl8192du

Understood. Given that the test environment, kernel module and test suite currently appear to be tn40xx specific, I think it still makes sense to keep it separate for now. Similar future (e.g. rtlX) network driver environments could reuse your scripts where needed (e.g. via `autorun/lib/`). 

> The goal for these drivers is probably to cease stand-alone development and be superseded by a mainline driver, but even then, the rapido developer workflow that enables passing the peripheral and driver into a VM is so much better than any of the alternatives I've seen (and not specific to network devices, or pci devices, or even device drivers).

Sounds like a worthy goal. Glad to hear that the rapido workflow is useful for others.

ddiss:
Thanks for the update, but this doesn't appear to address many of the previous comments:
* please rename it to match the specific functionality/driver to be tested, e.g. `cut/tn40xx.sh` and `autorun/tn40xx.sh`
* avoid adding new rapido.conf variables where possible
  * hardcode `NET_TEST_KMOD` as `tn40xx`
  * add a `TN40XX_DRIVER_SRC` variable, which is used instead of `NET_TEST_SUITE`
  * drop `NET_TEST_DEV_HOST`
    * instead you could check that `QEMU_EXTRA_ARGS` includes a `vfio-pci` device, similar to what we do in `cut/ocfs2.sh`
  * drop `NET_TEST_DEV_VM`: can this be hardcoded or discovered in the autorun script?

```
+       --include "${KERNEL_SRC}/samples/pktgen" "pktgen" \
+       --include "${NET_TEST_SUITE}" "driver-tests" \
```
Please use the host path in the VM:
```
+       --include "${KERNEL_SRC}/samples/pktgen" "${KERNEL_SRC}/samples/pktgen" \
+       --include "${TN40XX_DRIVER_SRC}/test" "${TN40XX_DRIVER_SRC}/test" \
```

```
+/driver-tests/start.sh ${NET_TEST_DEV_VM} ${NET_TEST_KMOD}
```
Depending on whether you think you'll only ever be running `start.sh`, you might want the autorun script to just have:
```
cd ${TN40XX_DRIVER_SRC}/test
```

you could then trigger the script via `./rapido cut -x "./start.sh DEV tn40xx" tn40xx`

acooks:
> Thanks for the update, but this doesn't appear to address many of the previous comments:
> 
>     * please rename it to match the specific functionality/driver to be tested, e.g. `cut/tn40xx.sh` and `autorun/tn40xx.sh`
> 
>     * avoid adding new rapido.conf variables where possible
>       
>       * hardcode `NET_TEST_KMOD` as `tn40xx`
>       * add a `TN40XX_DRIVER_SRC` variable, which is used instead of `NET_TEST_SUITE`
>       * drop `NET_TEST_DEV_HOST`
>         
>         * instead you could check that `QEMU_EXTRA_ARGS` includes a `vfio-pci` device, similar to what we do in `cut/ocfs2.sh`
>       * drop `NET_TEST_DEV_VM`: can this be hardcoded or discovered in the autorun script?
> 
> 
> ```
> +       --include "${KERNEL_SRC}/samples/pktgen" "pktgen" \
> +       --include "${NET_TEST_SUITE}" "driver-tests" \
> ```
> 
> Please use the host path in the VM:
> 
> ```
> +       --include "${KERNEL_SRC}/samples/pktgen" "${KERNEL_SRC}/samples/pktgen" \
> +       --include "${TN40XX_DRIVER_SRC}/test" "${TN40XX_DRIVER_SRC}/test" \
> ```
> 
> ```
> +/driver-tests/start.sh ${NET_TEST_DEV_VM} ${NET_TEST_KMOD}
> ```
> 
> Depending on whether you think you'll only ever be running `start.sh`, you might want the autorun script to just have:
> 
> ```
> cd ${TN40XX_DRIVER_SRC}/test
> ```
> 
> you could then trigger the script via `./rapido cut -x "./start.sh DEV tn40xx" tn40xx`

Thank you for the feedback. It's clear enough what you expect and I am still trying to incorporate it, but I'm not yet able to see a way to implement it in such a way that this software does what I need it to do.

With the changes in the PR I was able to do basic network device driver tests, like:
* loading and unloading the kernel module,
* setting the network interface up and down,
* changing MTU

These tests are obviously still quite basic and immature, but I couldn't find an existing reusable test suite and I need it, so I guess I'm building something.

The next tests I need to write will have to send and receive some actual network traffic and to do that, I'm working towards spinning up multiple VMs, passing a different PCI device with its appropriate driver to each VM from the host, and running the appropriate scripts to support support the test.

And that work will be balanced with working on the actual tn40xx driver.

I'll close this PR for now and perhaps try again when I have a better idea of how to get to an acceptable solution that still meets my needs.

Thanks again!

ddiss:
...
> Thank you for the feedback. It's clear enough what you expect and I am still trying to incorporate it, but I'm not yet able to see a way to implement it in such a way that this software does what I need it to do.
> 
> With the changes in the PR I was able to do basic network device driver tests, like:
> 
>     * loading and unloading the kernel module,
> 
>     * setting the network interface up and down,
> 
>     * changing MTU
> 
> 
> These tests are obviously still quite basic and immature, but I couldn't find an existing reusable test suite and I need it, so I guess I'm building something.
> 
> The next tests I need to write will have to send and receive some actual network traffic and to do that, I'm working towards spinning up multiple VMs, passing a different PCI device with its appropriate driver to each VM from the host, and running the appropriate scripts to support support the test.

Sounds good Andrew. FWIW, I'd still very much like to carry your changes upstream. I'd just prefer to keep things relatively self-contained and non-generic.

> And that work will be balanced with working on the actual tn40xx driver.
> 
> I'll close this PR for now and perhaps try again when I have a better idea of how to get to an acceptable solution that still meets my needs.

Please do try to resubmit when you have something somewhat self-contained that you're happy with.

PR #231, opened 2023-12-11
LTP: setup network early, add dmesg

pevik:


ddiss:
Merged - thanks Petr!

PR #232, opened 2024-01-02
LTP: Add mkfs.bcachefs support

pevik:


ddiss:
Pushed - thanks!
FWIW, we support pulling in user-space FS tools from `BCACHEFS_TOOLS_SRC`, `BTRFS_PROGS_SRC`, `EXFAT_PROGS_SRC`, etc. for the fstests runners. Might be worth supporting them in cut/ltp too if you think it'd be helpful.

pevik:
I was going to send this, but wouldn't be better to have `$BCACHEFS_TOOLS_SRC` + `PATH` modification in separate function and reuse it, e.g. add it to `runtime.vars` or `autorun/lib/fstests.sh`?

```diff
+++ autorun/ltp.sh
@@ -24,6 +24,11 @@ export KCONFIG_PATH=/.config
 export LTPROOT="$LTP_DIR"
 export PATH="$LTP_DIR:$LTP_DIR/bin:$LTP_DIR/testcases/bin:$PATH"
 
+if [ -n "$BCACHEFS_TOOLS_SRC" ]; then
+       export PATH="${PATH}:${BCACHEFS_TOOLS_SRC}"
+       ln -sfr $(type -P bcachefs) /bin/mkfs.bcachefs
+fi
+
 cd $LTP_DIR/testcases/bin/ || _fatal
 if [ -n "$LTP_AUTORUN_CMD" ]; then
        echo "Running LTP Command: $LTP_AUTORUN_CMD"

```

ddiss:
> I was going to send this, but wouldn't be better to have `$BCACHEFS_TOOLS_SRC` + `PATH` modification in separate function and reuse it, e.g. add it to `runtime.vars` or `autorun/lib/fstests.sh`?

We have `_rt_require_bcachefs_tools()`... if you want to share the symlink / PATH boot stuff between targets then we could do something similar to `_rt_write_ceph_config`, which generates pre-autorun commands in a temp file from the cut script.

PR #236, opened 2024-05-21
fstests: extract serial number for NVMe

hreinecke:
NVMe devices store their serial number in the 'device' sysfs directory, not the main block device sysfs directory.


ddiss (changes requested):
Thanks for the patch Hannes!

I might be missing something, but won't the nvme case be skipped over due to the prior `|| continue`?
(edit) nit: also, please use tabs for indentation

hreinecke:
Aw. Old patch. I'll be uploading the correct version.

ddiss (approved):
Looks good

PR #237, opened 2024-05-22
Fixup 'su -l' handling

hreinecke:
xfstests is calling 'su -' on several locations, but the invocation will fail with 'su: Critical error - immediate abort'.
Primary reason is a missing /etc/pam.d/su-l, but we also make sure the /etc/profile doesn't attempt to setup the system twice.

### DIFF
--- a/autorun/lib/fstests.sh
+++ b/autorun/lib/fstests.sh
@@ -93,5 +93,10 @@ auth	sufficient	pam_rootok.so
 account	sufficient	pam_rootok.so
 session	required	pam_limits.so
 EOF
+	cat > /etc/pam.d/su-l <<EOF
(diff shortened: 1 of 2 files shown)


ddiss:
Thanks Hannes. These look reasonable.
I think I used a different approach to "make script idempotent" in the past to address the re-run issue, but I'm not sure where it landed. I'll have a closer look and play with these changes over the next day or so.

ddiss:
`su` and device serial whitespace fixes were merged, the vm_autorun idempotence changes are no longer necessary following the change to use an explicit bash rcfile instead of `.profile` (see https://github.com/rapido-linux/rapido/pull/239 ).

Closing...

PR #239, opened 2024-05-28
autorun/init: start vm_autorun.env via bash --rcfile

ddiss:
Hanne's analysis:
  vm_autorun.env is copied into the rapido root fs as /etc/profile,
  and as such will be executed every time a login shell is invoked
  (ie via 'su -' or 'sh -l'). As this script also serves as system
  initialisation we need to make sure to run the initialisation only
  once, and only export the environment variables for later invocations.

To avoid multiple invocations, we can use bash's explicit --rcfile parameter when starting the initial shell, and install vm_autorun.env in a path where it won't get invoked automatically (/rapido.rc).

To use bash with --rcfile, we also need to drop the explicit --login/-l parameter, which changes bash behaviour slightly:
- (unused) profile paths are different
- legacy fcntl(fd=[3->19], F_SETFD, FD_CLOEXEC) calls are skipped

I *think* we shouldn't see any side-effects for rapido workloads.

Reported-by: Hannes Reinecke <hare@suse.de>

### DIFF
--- a/autorun/00-rapido-init.sh
+++ b/autorun/00-rapido-init.sh
@@ -1,15 +1,5 @@
-#
-# Copyright (C) SUSE LLC 2020, all rights reserved.
-#
-# This library is free software; you can redistribute it and/or modify it
-# under the terms of the GNU Lesser General Public License as published
-# by the Free Software Foundation; either version 2.1 of the License, or
-# (at your option) version 3.
-#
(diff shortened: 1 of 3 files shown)


ddiss:
I've been running with this change for a little while and haven't noticed any issues, so will go ahead and merge.

Issue #240, opened 2024-08-15
systemd-network 256.4-2.1 fails to bring up devices


ddiss:
Following a recent Tumbleweed update to `systemd-network-256.4-2.1`, I'm no longer seeing rapido network devices come up. IIRC, this was all working fine with `systemd-network-255.8-1.1`. Nothing stands out immediately from the strace.
(*edit:* I can bring the device up manually with `ip`, so the problem is unlikely to be kernel related)

Werkov:
systemd CI has been recently [oopsing in network setup code](https://github.com/systemd/systemd/actions/runs/10402701847/job/28807710342#step:12:504) when networkctl interacts with TW kernel. Do you spot any oopses in your case too?

ddiss:
> systemd CI has been recently [oopsing in network setup code](https://github.com/systemd/systemd/actions/runs/10402701847/job/28807710342#step:12:504) when networkctl interacts with TW kernel. Do you spot any oopses in your case too?

Interesting. No, I haven't seen anything like that; netlink I/O appears to be working fine. I'll take a closer look today...

ddiss:
Looks like udev interaction. On (broken) Tumbleweed I see:
```
eth0: Saved new link: ifindex=2, iftype=ETHER(1), kind=n/a
eth0: Saved original MTU 1500 (min: 68, max: 65535)
eth0: Saved permanent hardware address: b8:01:98:d6:7b:32
eth0: Saved hardware address: b8:01:98:d6:7b:32
eth0: Saved IPv6 link-local address generation mode: eui64
eth0: Flags change: +MULTICAST +BROADCAST
eth0: link pending udev initialization...
```

Leap 15.6 (working) logs:
```
Found link eth0(2)
eth0: Saved new link: ifindex=2, iftype=ETHER(1), kind=n/a
eth0: Saved original MTU 1500 (min: 68, max: 65535)
eth0: Saved permanent hardware address: b8:01:98:d6:7b:32
eth0: Saved hardware address: b8:01:98:d6:7b:32
eth0: Saved IPv6 link-local address generation mode: eui64
eth0: Flags change: +MULTICAST +BROADCAST
eth0: link has not yet been processed by udev: setup state is n/a.
eth0: udev initialized link
eth0: link has not yet been processed by udev: setup state is pending.
eth0: link has not yet been processed by udev: setup state is pending.
eth0: link has not yet been processed by udev: setup state is pending.
[...]

PR #241, opened 2024-08-19
fix network device bringup for systemd/udevadm v256

ddiss:
This fixes https://github.com/rapido-linux/rapido/issues/240 which was caused by an upstream systemd regression in https://github.com/systemd/systemd/pull/31998 . The fix also works with older systemd versions.

```
The following changes since commit 4149399fd4cbd6ad33be1264959da9e567ae0346:

  kernel: add sle15sp6_config (2024-07-31 17:31:23 +0200)

are available in the Git repository at:

  https://github.com/ddiss/rapido.git net_udevd

for you to fetch changes up to 363338a0bacf2d6d8017110cad79db100e2e6a32:

  kernel/vanilla_config: update to 6.11.0-rc4 (2024-08-19 05:58:56 +0000)

----------------------------------------------------------------
David Disseldorp (3):
      vm_autorun.env: start systemd-udevd prior to systemd-networkd
      cut/simple_network: add resize
      kernel/vanilla_config: update to 6.11.0-rc4

 cut/simple_network.sh |   2 +-
 kernel/vanilla_config | 257 ++++++++++++++++++++++++++++++++++++--------------
 runtime.vars          |  10 +-
 vm_autorun.env        |  14 +--
 4 files changed, 199 insertions(+), 84 deletions(-)
```


vm_autorun.env
```diff
@@ -182,16 +182,18 @@ _vm_ar_network_setup() {
 	       || _fatal "systemd.machine_id missing in kcli"
 	echo "$kcli_systemd_machine_id" > /etc/machine-id
 
-	for f in /sys/class/net/*; do
-		udevadm test $f > /dev/null 2>&1
-	done
+	PATH="/usr/lib/systemd:${PATH}"
+	systemd-udevd --daemon || _fatal
```
Werkov, line 186: I wonder how this would play along with setups that include `dracut-systemd` module. Maybe guard this under
```
if [[ $(cat /proc/1/comm) != systemd ]] ...
```

Actually, I see that's starting udevd is not a new idiom and the it's rather cuts with `dracut-systemd` that are novel, so the situation above I imagine never happens.

ddiss: > I wonder how this would play along with setups that include dracut-systemd module. Maybe guard this 

I've tried adding the `dracut-systemd` module to `simple-network` and it appears to run fine with manual invocation:
```
rapido1:/# systemctl status
● rapido1
    State: initializing
    Units: 81 loaded (incl. loaded aliases)
     Jobs: 1 queued
   Failed: 0 units
    Since: Mon 2024-08-19 09:58:21 UTC; 13s ago
  systemd: 256.4+suse.6.g5bba1ebe17
  Tainted: unmerged-bin
   CGroup: /
           ├─init.scope
           │ └─1 /init
           └─system.slice
             ├─dracut-cmdline.service
             │ ├─159 /bin/sh /bin/dracut-cmdline
             │ ├─209 /bin/bash --rcfile /rapido.rc -i
             │ ├─221 systemd-udevd --daemon
             │ ├─224 systemd-networkd
             │ ├─239 systemctl status
             │ └─240 less
             └─systemd-journald.service
               └─165 /usr/lib/systemd/systemd-journald
```

The image does carry a `systemd-udevd.service`, but it doesn't get triggered. I haven't looked closely as to why:
```
systemctl status systemd-udevd
○ systemd-udevd.service - Rule-based Manager for Device Events and Files
     Loaded: loaded (/usr/lib/systemd/system/systemd-udevd.service; static)
     Active: inactive (dead)
TriggeredBy: ○ systemd-udevd-control.socket
             ○ systemd-udevd-kernel.socket
       Docs: man:systemd-udevd.service(8)
```
Happy to accept changes to trigger this on systemd enabled boot if you'd like to have it :).

> Actually, I see that's starting udevd is not a new idiom and the it's rather cuts with dracut-systemd that are novel, so the situation above I imagine never happens.

`dracut-systemd` adds close to 10M to the `simple-network` image on Tumbleweed, so I'd like to continue to make it optional if we can.

ddiss:
I've merged this, as it's working for systemd and non-systemd images as is. Proper use of `systemd-udevd.service` with systemd can be handled later if desired.

PR #228, opened 2023-09-21
Distribution packaging changes

ddiss:
An attempt to address https://github.com/rapido-linux/rapido/issues/204 . Packages based on this RFC can be found at:
https://build.opensuse.org/package/show/home:dmdiss:rapido/rapido

```
The following changes since commit 7a71765ebcf2302962c7ec89432999f6400e73b0:

  autorun/ceph: load rbd module prior to sysfs IO (2023-09-11 12:31:33 +0200)

are available in the Git repository at:

  https://github.com/ddiss/rapido.git distro_pkg

for you to fetch changes up to 9eba94f75bca048a28b2d355060afd73f2e8edc0:

  tools/bash_completion: support rapido binary in PATH (2023-09-21 16:40:37 +0200)

----------------------------------------------------------------
David Disseldorp (4):
      build: add Makefile for install target
      build: generate and install /etc/rapido/rapido.conf
      runtime: use /etc/rapido/rapido.conf RAPIDO_CONF if present
      tools/bash_completion: support rapido binary in PATH

 Makefile              | 42 ++++++++++++++++++++++++++++++++++++++++++
 runtime.vars          |  8 ++++++--
 tools/bash_completion |  4 ++--
 3 files changed, 50 insertions(+), 4 deletions(-)
 create mode 100644 Makefile
```

PR #242, opened 2024-11-26
Preliminary distro packaging changes

ddiss:
The first minor batch of changes moving rapido towards distro package support, cherry picked from https://github.com/rapido-linux/rapido/pull/228 .
I've tacked on one extra selftests fix to handle upstream dracut output changes.

```
The following changes since commit c4e0cff00bfdaea03b960311266db5f59f7c5390:

  usbip_host_stub: cut and autorun scripts (2024-11-25 13:59:18 +1100)

are available in the Git repository at:

  https://github.com/ddiss/rapido.git prelim_distro_pkg

for you to fetch changes up to b33a5a1f5e80105bc5151b3827266d1a2cd9b78d:

  selftest/002: fix dracut-output needle (2024-11-26 15:52:33 +1100)

----------------------------------------------------------------
David Disseldorp (4):
      runtime: use /etc/rapido/rapido.conf RAPIDO_CONF if present
      tools/bash_completion: drop shebang
      rapido: don't cd into RAPIDO_DIR before cut invocation
      selftest/002: fix dracut-output needle

 rapido                | 18 ++++++------------
 runtime.vars          |  8 ++++++--
 selftest/test/002     |  2 +-
 tools/bash_completion |  1 -
 4 files changed, 13 insertions(+), 16 deletions(-)
```

### DIFF
--- a/rapido
+++ b/rapido
@@ -85,20 +85,17 @@ rapido_cut()
 		exit 1
 	fi
 
-	pushd "$RAPIDO_DIR" > /dev/null
-	cut_script="cut/${testname//-/_}.sh"
+	cut_script="${RAPIDO_DIR}/cut/${testname//-/_}.sh"
 
 	if [[ ! -x $cut_script ]]; then
 		[[ -f $cut_script ]] \
 		  && echo "$cut_script lacks execute permission." \
 		  || echo "$testname not found. See \"rapido list\"."
-		popd > /dev/null
 		exit
 	fi
 
-	./$cut_script "${post_autorun_files[@]}"
+	"$cut_script" "${post_autorun_files[@]}"
 	local cut_status=$?
-	popd > /dev/null
 	[ $cut_status -ne 0 ] && exit $cut_status
 	[ -n "$boot_img" ] || exit 0
 	"${RAPIDO_DIR}/vm.sh"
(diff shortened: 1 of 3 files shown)


runtime.vars
```diff
@@ -30,8 +30,12 @@ if [[ -n $RAPIDO_CONF ]]; then
 	# explicit user-provided conf path; fail if missing.
 	. "$RAPIDO_CONF" || _fail "$RAPIDO_CONF missing"
 else
-	# use default rapido.conf path; continue if missing.
-	RAPIDO_CONF="${RAPIDO_DIR}/rapido.conf"
+	# default conf path; continue if missing. etc path is for distributions
+	if [[ -f "/etc/rapido/rapido.conf" ]]; then
```
pevik, line 34: nit: maybe check also `/usr/etc/rapido/rapido.conf` (Micro?)

pevik (approved):
LGTM
```
```

runtime.vars
ddiss: Good point. With the Micro static (`/usr/etc/`) + dynamic (`/etc/`) config scheme it might be better to first work out whether merged configuration files would make more sense here instead. Will think about this and hold off on merging for now. Thanks for the review!

Werkov: User anecdote: my rapido configs are almost always different (~per project). I'm not sure, I'd make use of composition/stacking or even the global config in `/etc`. To me the file plays a role it's more like an extra input to each rapido call, not a persistent/global config. That'd mean for packaging: a) fail without the config, b) fall back to a set default values.

ddiss: I'll drop the global config change for now, as it's easier to add later rather than remove when (mostly) unused. For distro packages we could just expect a `RAPIDO_CONF` env for any non-default settings.

PR #244, opened 2025-01-29
Make /usr writable in /usr images with systemd (again)

Werkov:
systemd v256 made /usr inside initrds immutable with RO-mount but this may be undesired for rapido test images with systemd. Let's ship a systemd drop-in that disables this /usr protection.

I hesitate with this workaround since it adds the drop-in unconditionally whereas it's only needed with `dracut-systemd` module, I don't know how to filter that.

### DIFF
--- a/runtime.vars
+++ b/runtime.vars
@@ -273,7 +273,8 @@ _rt_require_dracut_args() {
 	[[ -f $RAPIDO_CONF ]] || conf_src="${RAPIDO_DIR}/dracut.conf.d/.empty"
 	DRACUT_RAPIDO_ARGS+=(--include "$conf_src" /rapido.conf \
 			     --include "$env_src" /rapido.rc \
-			     --include "$rinit_src" "$rinit_dst")
+			     --include "$rinit_src" "$rinit_dst"\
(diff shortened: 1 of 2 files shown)


ddiss (approved):
Looks good to me, thanks Michal! Will have a play with this over the next day or so and then merge.

ddiss:
I had a closer look at the upstream systemd changes (`config_parse_protect_system_pid1()` and `apply_protect_system()`). I now agree that this probably should only be applied if the corresponding `autorun` script actually needs to modify `/usr`.
It should be relatively easy to generate a tmp `system.conf.d` file as part of the `cut` script, e.g. https://github.com/rapido-linux/rapido/blob/a4fa16f4aae859b2dee0492de7eaf7614dd3ed73/cut/uring_btrfs.sh#L12 .

Of the few systemd-enabled runners we have so far, it looks like `lrbd` is the only one that attempts to modify under `/usr`. I'll add the config-gen logic there.

ddiss:
> ...
> Of the few systemd-enabled runners we have so far, it looks like `lrbd` is the only one that attempts to modify under `/usr`. I'll add the config-gen logic there.

Proposed via https://github.com/rapido-linux/rapido/pull/245

PR #245, opened 2025-01-30
lrbd: make /usr writable with systemd (again)

ddiss:
systemd v256 made /usr inside initrds immutable with RO-mount, while autorun/lrbd.sh attempts to create some python symlinks there. Generate and install a systemd conf that disables this /usr protection.

I don't have a Ceph+lrbd environment available for testing, but this should still help as an example for other systemd-based runners.


Werkov (approved):
LGTM, let's stick to static `/usr` then.

PR #249, opened 2025-06-11
lkl: remount /dev/shm without noexec

ddiss:
LKL MMU=y mode calls mmap() with PROT_EXEC, so the Dracut noexec default causes LKL boot failure if retained.

Also add a few extra mkfs binaries for disk.sh -t $TYPE testing.


Werkov:
LGTM.

Issue #254, opened 2025-08-08
selftests fail due to SUSE Dracut default log level change


ddiss:
Tumbleweed appears to be carrying the following Dracut change, which breaks rapido `selftests`:
```
commit 311f84f698fb333fa80af4fe6b2a78304b4364cb
Author: Antonio Alvarez Feijoo <antonio.feijoo@suse.com>
Date:   Thu Sep 5 13:46:35 2024 +0200

    ci(suse.conf.example): change log levels
    
    Reduce default standard log verbosity from `info` to `warning`, and increase
    syslog verbosity from none to `info`.
    
    jsc#PED-12922

diff --git a/dracut.conf.d/suse.conf.example b/dracut.conf.d/suse.conf.example
index b9b0c372..aed4d432 100644
--- a/dracut.conf.d/suse.conf.example
+++ b/dracut.conf.d/suse.conf.example
@@ -12,3 +12,5 @@ compress="zstd -3 -T0 -q"
 i18n_vars="/etc/sysconfig/language:RC_LANG-LANG,RC_LC_ALL-LC_ALL /etc/sysconfig/console:CONSOLE_UNICODEMAP-FONT_UNIMAP,CONSOLE_FONT-FONT,CONSOLE_SCREENMAP-FONT_MAP /etc/sysconfig/keyboard:KEYTABLE-KEYMAP"
 omit_drivers+=" i2o_scsi "
 
+stdloglvl="3"
+sysloglvl="4"
```

Adding `stdloglvl="4"` to `dracut.conf.d/01-rapido-dracut.conf` appears to be all that's needed to get things working again.
The tests should probably be fixed to avoid relying on any Dracut output, but let's go with a config change for now.

ddiss:
If you have a more portable approach to determining where the host kernel lives, then I'm all ears :) . For now it should work if you export the `KERNEL_SRC` and `KERNEL_INSTALL_MOD_PATH` env vars.

PR #255, opened 2025-09-02
Add CI with selftests in a container

Werkov:
Followup to [this](https://github.com/rapido-linux/rapido/issues/254#issuecomment-3175282786)

One could install `kernel-default-devel` and `KERNEL_SRC=/usr/src/linux-6.16.3-1-obj/x86_64/default/` + tweak lookup of vmlinuz. Or teach rapido to use specific kernel version (commit in this PR) and then run the tests in a simple container image:

```Dockerfile
FROM registry.opensuse.org/opensuse/tumbleweed:latest
RUN zypper --non-interactive in dracut expect kernel-default 
RUN zypper --non-interactive in qemu-x86

# only needed with KERNEL_SRC override, won't work without further changes in rapido
# RUN zypper --non-interactive in kernel-source
# RUN zypper --non-interactive in kernel-devel
# RUN zypper --non-interactive in kernel-syms

CMD /test/selftest/selftest.sh
```

And then I can run selftests like:
```
docker run -v $PWD:/test -w /test \
    -e KERNEL_RELEASE=6.16.3-1-default
    -e KERNEL_INSTALL_MOD_PATH=/usr/
    localhost/rapido-test-image
```

### Next up:

- push the image somewhere
- [figure out how to make Github CI](https://stackoverflow.com/a/64200200) run actions inside said image
- possibly automate building/refreshing the image + automatic updated of KERNEL_RELEASE used in CI

### DIFF
--- a/rapido.conf.example
+++ b/rapido.conf.example
@@ -1,10 +1,17 @@
 # Path to Linux kernel source. Prior to running "rapido cut", the kernel
 # should be built, with modules installed (see KERNEL_INSTALL_MOD_PATH below).
-# If KERNEL_SRC is not set then rapido will attempt to boot the local kernel
-# present under /boot/.
+# If KERNEL_SRC is not set then rapido will attempt to boot a host kernel
+# available under /boot/; either the `uname -r` release, or (if specified) the
+# KERNEL_RELEASE version.
 #
 # e.g. KERNEL_SRC="/home/me/linux"
 #KERNEL_SRC=""
+
+# Only applicable with an empty KERNEL_SRC. KERNEL_RELEASE can be used to
+# override which host kernel should be used, when left empty/unset defaults to
+# `uname -r`.
+# e.g. KERNEL_RELEASE="6.16.3-1-default"
+#KERNEL_RELEASE=""
 
 # If specified, this parameter defines the path that Dracut should use to
 # obtain compiled kernel modules. If left blank, Dracut will use its default
--- a/runtime.vars
+++ b/runtime.vars
@@ -308,6 +308,8 @@ _rt_require_dracut_args() {
 		kver="$(cat "${KERNEL_SRC}/include/config/kernel.release")"
 		[[ -n $kver ]] \
 			|| _fail "failed to read kernel.release at $KERNEL_SRC"
+	elif [[ -n $KERNEL_RELEASE ]] ; then
+		kver="$KERNEL_RELEASE"
 	else
 		kver="$(uname -r)"
 	fi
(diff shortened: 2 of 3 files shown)


ddiss:
The new setting looks reasonable to me. Please document its use in `rapido.conf.example`. The same setting should be usable for "regular" Linux hosts with multiple kernel versions installed, right?

thackara:
if we had installed `kernel-default-devel`, the `KERNEL_RELEASE` can be read from filesystem rather than an ENV?
```
# cat /usr/src/linux-obj/x86_64/default/include/config/kernel.release
6.16.7-1-default
```

Werkov:
I made it work in my fork of the repo (where I commited `.github/workflows/` into `master`. [See](https://github.com/Werkov/rapido/actions/runs/17864573572/job/50803653761?pr=1#step:4:506).

The test image is built and hosted in [OBS](https://build.opensuse.org/package/show/home:mkoutny:containers/rapido-ci). Now, I installed all with 'latest' version but then I have to put absolute number in `ci.yml` `KERNEL_RELEASE`. That would work if the container image was never rebuilt.

I may look into the suggestion with `kernel-default-devel` and filling it automatically inside this test container from `ci.yml`. (Although, I'm afraid it would not generalize for generic rapido use with any distro as `-devel` is openSUSE specific.)

Unmarking draft for more feedback.

rapido.conf.example
```diff
@@ -1,11 +1,16 @@
 # Path to Linux kernel source. Prior to running "rapido cut", the kernel
 # should be built, with modules installed (see KERNEL_INSTALL_MOD_PATH below).
-# If KERNEL_SRC is not set then rapido will attempt to boot the local kernel
-# present under /boot/.
+# If KERNEL_SRC is not set then rapido will attempt to boot the local running
+# kernel present under /boot/ unless KERNEL_RELEASE is set.
```
ddiss, line 4: I think I'd prefer:
```
# If KERNEL_SRC is not set then rapido will attempt to boot a host kernel
# available under /boot/. Either the `uname` release, or the version explicitly specified via KERNEL_RELEASE
```

ddiss (changes requested):

Werkov:
Added `kernel-default-devel` into the container image for autodetection of `KERNEL_RELEASE `in CI. Addressed feedback on comments. CI restared in a [user fork](https://github.com/Werkov/rapido/pull/1).

runtime.vars
```diff
@@ -308,6 +308,8 @@ _rt_require_dracut_args() {
 		kver="$(cat "${KERNEL_SRC}/include/config/kernel.release")"
 		[[ -n $kver ]] \
 			|| _fail "failed to read kernel.release at $KERNEL_SRC"
+	elif [[ -n $KERNEL_RELEASE ]] ; then
+		kver=$KERNEL_RELEASE
```
ddiss, line 312: The assignment here should be quoted. I'll add that when I merge.

ddiss (approved):
Looks good, thanks! Will merge with the minor quoting change squashed in.

PR #258, opened 2025-12-03
add native rust-based initramfs generator, init and qemu wrapper

ddiss:
Sangeetha and I have been working on some changes to replace Dracut with
a rust-based initramfs / cpio image generator. The main reasons for this
are:
- Some distros are moving away from using Dracut, so we can't assume it's
  locally available
- We only use a small portion of Dracut functionality: base and systemd
  modules, and kernel / user dependency gathering for cpio
- Dracut is slow: it forks many processes, stages all initramfs content
  and is mostly written in bash

The rewrite builds on my previous dracut-cpio implementation and adds:
- elf dependency gathering using https://github.com/cole14/rust-elf
- kernel module dependency gathering via native modules.dep (etc.)
  parsing (thanks @thackara !)
  - VMs still user regular kmod / modprobe.
- basic rapido-vm and rapido-init programs, to run qemu and start autorun
  scripts
- basic rapido.conf parsing

It's otherwise kept as minimal as possible, with rust-elf and the std
library the only major external dependencies. One single-file crosvm
argument parser is also bundled.

The preliminary benchmark results look good, particularly for initramfs
image generation (cut):
```
---------------+-------------------------------+-----------------------|
               | Before: rapido e4c6077        | After: rs_wip 9a90973 |
               | dracut-059+suse.769.g693ea004 |  rustc 1.91.0         |
---------------+-------------------------------+-----------------------|
simple-example |   2.389s +- 0.117             | 0.075s +- 0.001       |
cut            |                               |                       |
---------------+-------------------------------+-----------------------|
simple-example |   7.2942 +- 0.065             | 4.746s +- 0.003       |
cut+boot+exit  |                               |                       |
---------------+-------------------------------+-----------------------|
simple-network |   2.572s +- 0.135             | 0.098s +- 0.000       |
cut            |                               |                       |
---------------+-------------------------------+-----------------------|
simple-network |   7.460s +- 0.126             | 4.926s +- 0.011       |
cut+boot+exit  |                               |                       |
---------------+-------------------------------+-----------------------|
```

To try these changes yourself, check out this branch and run:
```
cargo build --offline --release
./rapido cut simple-example
```
You may need to change your rapido.conf file a little if you
use env variables or shell callouts. 

I'm flagging this as WIP, as there are still a few things to do:
- boot with systemd as init, instead of only rapido-init
- clean up interface before locking it in (parameter naming, etc.)
- think about distro packaging / path assumptions
  - at the moment, it assumes bins are in target/release/* and conf is
    in the working directory
- improve test coverage

I don't expect to convert remaining cut scripts before merge. Dracut and
rust based functionality should be able to live side by side, although
rapido.conf parsing is much less flexible in rust: no invocations,
currently no env var expansion, variables must be wrapped in {}.


src/bin/kmod/kmod_context.rs
```diff
@@ -39,7 +39,7 @@ pub struct KmodModule {
     pub status: ModuleStatus,
     // rel_path is relative to KmodContext.module_root
     pub rel_path: PathBuf,
-    pub hard_deps: Vec<String>,
+    pub hard_deps_paths: Vec<String>,
```
thackara, line 42: [minor]
or maybe: `pub hard_deps_paths: Vec<PathBuf>`  and later in hard_deps()->.map(|p| extract_module_name(p.rel_path.as_os_str().to_str().unwrap()))

-> Provides direct|semantic relationship with the data stored there.

[EDIT after reading the commit-msg]
this is the logic to "keep paths and generate hard_deps on demand", so your approach makes sense.

ddiss: Thanks for the review! PathBuf does feel more suitable but I don't think it is, as BufReader.lines() already ensures that we're working with UTF-8 strings. Also, I don't expect the kernel to be shipping non-UTF8 named kernel modules in future :) .
For consistency with `rel_path`, I'll push an extra change which switches that over to String.

pevik:
> I don't expect to convert remaining cut scripts before merge. Dracut and
rust based functionality should be able to live side by side, although
rapido.conf parsing is much less flexible in rust: no invocations,
currently no env var expansion, variables must be wrapped in {}.

What a huge rewrite (hackweek :)). I like at least some of the shell scripts got rewritten into rust.

ddiss:
I think these changes are pretty much ready for inclusion now, with systemd init support the only blocker.
The new manifest format is quite flexible, which should make conversion from dracut based "cut" scripts relatively straightforward. [simple-example](https://raw.githubusercontent.com/ddiss/rapido/refs/heads/rs_wip/cut/simple_example.sh), [simple-network](https://raw.githubusercontent.com/ddiss/rapido/refs/heads/rs_wip/cut/simple_network.sh) and [fstests-btrfs](https://raw.githubusercontent.com/ddiss/rapido/refs/heads/rs_wip/cut/fstests_btrfs.sh) are converted to rapido-cut, with `simple-systemd` still to come.

Existing dracut scripts will continue to work here, but use rapido-init and rapido-vm (as qemu wrapper) for boot. I'll branch before merge so that people can continue to use the unmodified Dracut boot path.

ddiss:
Basic systemd boot is now working for me on Leap 15.6, ubuntu-latest (Github CI) and Tumbleweed.

@Werkov it'd be great if you could have a play with the converted `kselftest` runner to see if anything is missing.
As a motivator, here are the `perf stat --repeat 10 ./rapido cut -B kselftest` benchmark numbers from my old laptop...
Before (current master branch with Dracut): 2.9523 +- 0.0380 seconds
After (rs_wip branch rapido-cut.rs): 0.037130 +- 0.000707 seconds

I plan on merging this very soon. I'll branch current master beforehand so that the Dracut based functionality remains easily accessible.

PR #266, opened 2026-05-14
kmod: fix builtin module soft dependency tracking

ddiss:
Sangeetha reported that a built-in module will not correctly track any soft dependencies. modules.builtin is processed after modules.softdep, while modules.builtin processing inserts hash entries and modules.softdep processing only modifies existing entries.

Fix this by processing modules.builtin before modules.softdep.

Reported-by: sangeetha.thackarajan@suse.com


thackara:
looks good to me!

pevik (commented):
LGTM.
```
```

ddiss:
Merged - thanks for the reviews!

PR #269, opened 2026-05-19
ci: switch from ubuntu-latest to ubuntu-slim

ddiss:
The ubuntu-latest runner provides 16GB RAM, which we don't need given that we're no longer running xfstests.


ddiss:
`ubuntu-slim` is unfortunately not an option for various reasons:
- no kvm
- no ability to create tap network devices
- no access to running kernel

It looks as though it runs as an unprivileged container.

